PR #27, opened 2022-05-13
even more headers

Cvolton:
The 10 or so cells with updateBGColor pointing to the same offset are not a copy paste error, in these cases it really is a different function on Android but it's all merged into one on Windows because all of these execute the exact same code

Delegates in GLM are taken from Wylie's GD decomp and verified to be mostly accurate (I didn't test every single one but I did a quick test of about a half of them)

also this one actually compiles which i think is an improvement from the last pr i made


HJfod:
fantastic pull request, pleasure doing business with you

PR #28, opened 2022-05-13
improved gamemanager

Cvolton:
this addresses any crashes caused by accessing the 3 dicts in GM and apart from fixing these being misaligned, it also fixes the misalignment of the 2 members that follow m_smoothFix

idk if this works on mac i just made the same edits to the mac offsets that i did for the windows ones

m_adTimer, m_adCache and m_unknownDouble might (still?) be misaligned, these were not tested, as i do not have ads in the steam version of geometry dash


Cvolton:
my new hobby is pull request spam

PR #36, opened 2022-08-14
IDK figure it out

SergeyMC9730:
Geode is not supporting Android yet. Please wait, before Android support will be eventually out.


matcool:
Not as funny of a msg, + L

SergeyMC9730:
Ratio

PR #102, opened 2022-12-24
Hooks

MikaKC:
Adds some other functions to hook in GeometryDash.bro (Windows only)


altalk23:
why does it look like all of the file has changed uh

MikaKC:
i just uploaded the whole file using GitHub

PR #109, opened 2023-01-09
Added the supported metering variable to PlayLayer

SMJSGaming:
From my findings, legacy levels are indicated if this bool is false. If it is, GD will default to extracting the metering from the AudioEffectsLayer rather than using GameSoundManager.

I say this is likely a way of indicating a legacy mode because this can be found in all levels which were made before 2.1.


SMJSGaming:
Quickly adding this for documentation sake. The red dashes lines are this calls in PlayLayer::updateVisibility

image

PR #141, opened 2023-03-05
Fix build if cli path contains spaces

cgytrus:


matcool:
is this really needed? cmake should already deal with it
(this is without the pr changes)
image

HJfod:
image

matcool:
cant repro.. theres also like 10 other uses of ${GEODE_CLI} like that if you really wanted to fix it

HJfod:
I think this is worth merging since it fixes an issue someone had and doesn't break other people's stuff

HJfod:
If this PR fixed Config's issue then yeah apparently

cgytrus:
i have no idea why it breaks in specifically that place specifically for me but adding those quotes indeed fixes the issue for m- what the hell it's working now???? nevermind then i guess, sorry lmao

PR #144, opened 2023-03-10
Fix CCDictionaryExt's `=` operator

SergeyMC9730:
that was easy and working fix


matcool:
how did this even compile before.. silly c++

PR #149, opened 2023-03-31
String IDs for CustomSongWidget in LevelInfoLayer

Fleeym:
I need those string IDs for a mod I'm creating so I guess I should make a pr for them too.

I'm actually not sure if I did this correctly.

### DIFF
--- a/loader/src/loader/Event.cpp
+++ b/loader/src/loader/Event.cpp
@@ -4,7 +4,9 @@
 using namespace geode::prelude;
 
 void EventListenerProtocol::enable() {
-    Event::listeners().push_back(this);
+    if (!ranges::contains(Event::listeners(), this)) {
+        Event::listeners().push_back(this);
+    }
 }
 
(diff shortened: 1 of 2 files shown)


HJfod:
These IDs should be set in the `CustomSongWidget` class itself and not in `LevelInfoLayer`, make a new file `loader/src/ids/CustomSongWidget.cpp` and hook its init to add the IDs there

Fleeym:
One small problem, CustomSongWidget::init isn't in the bindings

HJfod:
You can add it, its address is `0x685b0` on Windows

Fleeym:
Alright, done

HJfod:
@altalk23 Can you get the MacOS address for `CustomSongWidget::init`? Still gotta add that before merging so the commit can build

altalk23:
its `0x37be20`

Fleeym:
oops i broke something, i'll open another pr

PR #152, opened 2023-04-04
init binding and string ids for CustomSongWidget

Fleeym:
Did some commits with another email, oops. I added the mac binding.


altalk23 (approved):
looks good to me

PR #166, opened 2023-05-07
claim/unclaimParticle symbols

DexterHill0:
`PlayLayer::claimParticle` and `PlayLayer::unclaimParticle` were flipped. `PlayLayer::claimParticle` also returns the particle and not void.


matcool:
can you remove the comment, its outdatede

matcool:
also why did you flip the mac address, is it also wrong?

matcool:
just checked, mac addresses were correct, so flip those around too :)

DexterHill0:
i am confused on what the mac addresses should be now i didnt touch the addressed i just flipped the function names

matcool:
yes you switched the function names and args but kept the addresses, which now makes the mac address incorrect

just keep it as it was before and flip the windows address, and change the return type for claimParticle

DexterHill0:
ok i changed them are you sure theyre fine on mac

matcool:
yes they are fine on mac

PR #182, opened 2023-06-01
Fix Mod::setSettingValue

cgytrus:
this should be `key` instead of `sett`


altalk23:
Screen Shot 2023-06-04 at 17 50 45

PR #183, opened 2023-06-01
Fix GLFW headers

cgytrus:
The GLFW version there currently is (3.0.4) is not the correct GLFW version (3.1)
This PR fixes the headers and removes the glfw3.lib which is useless because all the glfw state is in the cocos library because its linked statically


matcool:
This PR has been approved by the GLFW Experts Working Group

PR #187, opened 2023-06-10
Update GeometryDash.bro

ToasterCoder44:


ToasterCoder44:
they sometimes get some weird values, but they're in most cases collided object unique ids

PR #216, opened 2023-08-22
change separator length

MacCraker:
the first 2 separators containing one more "=" character than the rest wasn't being very kind to my ocd so I had to change it, I know it's like the most unimportant change ever but still


altalk23:
cat gif

PR #266, opened 2023-09-25
Add CCTransitionFade::initWithDuration for mac

ninXout:


Cvolton (approved):
I think this is a high quality commit from a high quality pull request and should be accepted as soon as possible so ninXout from Crystal Client can release Crystal Client 4.0.2. In fact I believe the review process should be sped up from its usual procedures, so we do not risk ninXout starting work on Crystal Client 5.0.0 before this pull request gets accepted.

cgytrus:
I think this is a high quality commit from a high quality pull request and should be accepted as soon as possible so ninXout from Crystal Client can release Crystal Client 4.0.2. In fact I believe the review process should be sped up from its usual procedures, so we do not risk ninXout starting work on Crystal Client 5.0.0 before this pull request gets accepted.

matcool:
this commit sucks, dislike,

altalk23:
I think this is a low quality commit from a low quality pull request and should not be accepted as soon as possible so ninXout from Crystal Client can not release Crystal Client 4.0.2. In fact I do not believe the review process should be sped up from its usual procedures, so we risk ninXout starting work on Crystal Client 5.0.0 before this pull request gets accepted.

PR #267, opened 2023-09-26
Update Bindings for a bunch of classes (Mac)

FireMario211:
This PR adds mac offsets to existing functions. The classes include `SecretLayer2`, `LevelInfoLayer`, and `FriendsProfilePage`

If there needs to be any changes to this PR such as signatures or issues, let me know so I can make them.


altalk23:
I think this PR requires quite a lot of signature changes as it does not adhere with our Geode Contributing guidelines.

PR #268, opened 2023-09-26
add GJAccountManager::getGJP

dankmeme01:
i didn't find either of those but thanks Coloride for win and Firee for mac


altalk23:
people will now steal passwords using gjp nooooooo !!!!!!!!!!!!

PR #276, opened 2023-10-01
Add a lot of mac bindings

FireMario211:
This PR adds a lot of mac offsets to 2 classes, multiple in `CreatorLayer`, and one in `DailyLevelPage` as it was missing one.

If there needs to be any changes to this PR, let me know so I can make them.


matcool:
these are definitely some bindings

PR #278, opened 2023-10-01
Added a filter to only test offsets on bindings

SMJSGaming:
This way you prevent useless offset tests when there were no changes made to the bindings


matcool:
can you also add `loader/test/members`

PR #308, opened 2023-10-16
Bindings: Fix DailyLevelPage updateTimers calling convention

Cvolton:


altalk23:
contrary to popular belief, this binding in fact does more than adding a single WORD, since it adds an entire LPCHAR

PR #319, opened 2023-10-23
Display messages of unhandled c++ exceptions on Windows

dankmeme01:
When a C++ exception was thrown from a mod or from Geometry Dash itself, rather than displaying the error code `0xE06D7363`, attempt to also display the error message of the exception.


altalk23:
idk it probably works its windows only anyway

PR #321, opened 2023-10-27
Handle HTTP error responses

FireMario211:
By default, when an HTTP error code is sent (4XX or 5XX), this will usually result in the message `Fetch failed: HTTP response code said error`

To resolve this, FAILONERROR for curl opts is disabled, and HTTP error codes are handled manually. Instead of the generic response given before (Which doesn't tell anything about what really happened), the response is given instead.

If there needs to be any changes to this PR, let me know so I can make them.


dankmeme01 (approved):

altalk23:
gg this will prob not break anything so go

PR #368, opened 2023-12-10
Add bindings for GDL

Kolyah35:
AchievementBar members from gd.h (works perfectly)

I decompiled GauntletNode, GauntletSelectLayer and MultilineBitmapFont myself (it seems to be working fine)

### DIFF
--- a/bindings/GeometryDash.bro
+++ b/bindings/GeometryDash.bro
@@ -4647,7 +4647,8 @@ class MultilineBitmapFont : cocos2d::CCSprite {
 	gd::string stringWithMaxWidth(gd::string str, float scaledWidth, float scale) = win 0x2b450;
 	bool initWithFont(const char* fontName, gd::string str, float scale, float width, cocos2d::CCPoint anchorPoint, int unk, bool bColourDisabled) = win 0x2a6a0;
 
-	float m_fontWidth[300];
+	float m_fontWidth;
+    PAD = win 0x4AC, android 0x4AC;
 	cocos2d::CCArray* m_colouredTextArray;  //ColoredSection*
 	cocos2d::CCArray* m_instantTextArray;   //InstantSection*
 	cocos2d::CCArray* m_delayedTextArray;   //DelayedSection*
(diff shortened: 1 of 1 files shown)


bindings/GeometryDash.bro
```diff
@@ -4623,7 +4639,27 @@ class MoreVideoOptionsLayer : FLAlertLayer {
...
-class MultilineBitmapFont : cocos2d::CCSprite {}
+class MultilineBitmapFont : cocos2d::CCSprite {
+	static MultilineBitmapFont* create() = win 0x2a440;
+	static MultilineBitmapFont* createWithFont(const char* fontName, gd::string str, float scale, float width, cocos2d::CCPoint anchorPoint, int unk, bool bColourDisabled) = win 0x2a5b0;
+	gd::string readColorInfo(gd::string str) = win 0x2afb0;
+	void setOpacity(int) = win 0x2af50;
+	gd::string stringWithMaxWidth(gd::string str, float scaledWidth, float scale) = win 0x2b450;
+	bool initWithFont(const char* fontName, gd::string str, float scale, float width, cocos2d::CCPoint anchorPoint, int unk, bool bColourDisabled) = win 0x2a6a0;
+
+	float m_fontWidth[300];
```
altalk23, line 4650: id say using std::array<float, 300> m_fontWidth; would be more preferable

altalk23: and probably a better name, since floatwidth implies a single value, not an array

bindings/GeometryDash.bro
```diff
@@ -5,6 +5,7 @@ class GDString {
     GDString& winAssign(GDString const&, size_t, size_t) = win 0xf720;
     GDString& winAssign(char const*) = win 0xf680;
     GDString& winAssign(char const*, size_t) = win 0xf840;
+	GDString& operatorPlus(char const*, size_t) = win 0x156F0;
```
altalk23, line 8: not sure why this was added since its just an unused internal function but you go girl

bindings/GeometryDash.bro
Kolyah35: I think this is a mistake by the gd.h developers, there shouldn't be an array here.

PR #371, opened 2023-12-13
Fix Android strip not being run without Geode CLI

Cvolton:


Cvolton:
Using the parameter `-DGEODE_DISABLE_CLI_CALLS=1` while configuring causes the strip step for the Android binary to get skipped. Since the binary itself changing depending on this argument doesn't appear to be intended behavior, this PR fixes the issue

PR #376, opened 2023-12-17
Add LevelSelectLayer::colorForPage

TheSillyDoggo:


Capeling:
W

bindings/GeometryDash.bro
```diff
@@ -4174,6 +4174,7 @@ class LevelSearchLayer : cocos2d::CCLayer {
 class LevelSelectLayer : cocos2d::CCLayer {
     static LevelSelectLayer* create(int lvl) = win 0x185500;
     bool init(int lvl) = win 0x1855a0, mac 0x2384e0;
+    static uint colorForPage(int page) = win 0x18aaa0;
```
matcool, line 4177: why is this uint. how did this even compile. is this a color value or an index? if its a color value its most likely a `cocos2d::ccColor3B`

TheSillyDoggo:
Gjira said uint

matcool:
not only is ghidra guessing the return type, but its also pseudocode, containing other made up types such as undefined2 undefined4 etc. in this case uint is not made up but its just not what its called in c/c++

TheSillyDoggo:
Ok I didn't know that I've just learned about re'ing this week

Capeling:
this is a `cocos2d::ccColor3B`

PR #386, opened 2023-12-25
Refactor logging internals

### LINKED ISSUE #369: Refactor logging system
### Geode Version

v2.0.0

### Related Problem

The internal code for the logging system is overly complicated, and doesn't allow for any custom format specifiers.
The reason for its complexity was to integrate with DevTools and potentially other custom consoles. However, this was never implemented, and imo isnt that needed

### Your Solution

Just pipe to fmtlib directly or something

### Additional Information

_No response_

matcool:
Fixes #369 
- removes LogComponents, now logs just get formatted into a string and saved like that
- Now to define formatting of own type, use fmtlib's api (either format_as or fmt::formatter, see their docs!)
- Breaks ABI (wowie!)
- Usage wise, your code stays the exact same. unless you had defined your own `parser` overload

# TODO
- [x] Refactor Logger class?
- [x] ~~Make Log pimpl?~~ was made hidden instead

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,7 +81,7 @@ if (PROJECT_IS_TOP_LEVEL AND NOT GEODE_BUILDING_DOCS)
 	set(MAT_JSON_AS_INTERFACE ON)
 endif()
 CPMAddPackage("gh:geode-sdk/json#a47f570")
-CPMAddPackage("gh:fmtlib/fmt#9.1.0")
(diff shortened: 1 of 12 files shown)


matcool:
The log file names and msgs aren't using the local time (its always displaying utc time), i couldnt figure it out in time

matcool:
Fixed now, along with formatting Mod*, so everything should be done now :-)

PR #394, opened 2023-12-28
Update some enums to 2.2

Cvolton:


Cvolton:
dunno what to call search types 19 and 26 exactly btw, theyre valid but unused in the client - both of them accept a list of up to 100 levels separated by commas in str and the difference is that 19 splits it into 10 pages and 26 dumps all results onto a single one. both are assumed to have been used for old versions of lists but there's no real way to confirm this

PR #401, opened 2023-12-31
Version check to prevent loading in not 2.2 (windows only)

TheSillyDoggo:
og code by absolute modified by me (it didn't compile for me)
the gd versions are in loader/src/VersionDetect.h


Fleeym:
nice commit history

matcool:
Have you even tested this? I'm pretty sure this won't even work.
The xinput dll has geode.dll as a native dependency, so geode will be loaded before we can run any code. 
Also the code doesnt match our code style at all

TheSillyDoggo:
just remove it then.

PR #453, opened 2024-01-24
Launch Argument Additions

acikek:
## Additions

- **Safe Mode**. Can be enabled with `--geode:safe-mode`. Completely skips the mod loading step.
- **Disable Custom Menu**. Can be toggled on with `--geode:disable-custom-menu`.
  - Removes text below loading bar on startup
  - Removes "updating mod index" message
  - Removes Geode button in main menu
  - **Does not** disable any Geode popups/notifications from errors occurring
- `std::optional<T> parseLaunchArgument(std::string_view const name)` on both `Loader` and `Mod`. Makes it easier to parse integer, double, etc. launch arguments, and uses `matjson` for the deserialization
  - Renamed `getLaunchBool` to `getLaunchFlag` in light of this change, but that's fine since no one's using it anyways.

> **Note**: this also establishes a `kebab-case` naming convention for launch arguments.


loader/include/Geode/loader/Loader.hpp
```diff
@@ -94,17 +97,40 @@ namespace geode {
...
          */
         std::optional<std::string> getLaunchArgument(std::string_view const name) const;
         /**
-         * Get a boolean launch argument. Returns whether the argument is present and its
+         * Get a launch argument flag. Returns whether the argument is present and its
          * value is exactly `true`.
          * @param name The argument name
          */
-        bool getLaunchBool(std::string_view const name) const;
+        bool getLaunchFlag(std::string_view const name) const;
```
altalk23, line 110: im worried about this causing undefined symbol errors in newer mods

loader/include/Geode/loader/Mod.hpp
```diff
@@ -163,17 +168,26 @@ namespace geode {
...
          */
         std::optional<std::string> getLaunchArgument(std::string_view const name) const;
         /**
-         * Equivalent to a prefixed `Loader::getLaunchBool` call. See `Mod::getLaunchArgument`
+         * Equivalent to a prefixed `Loader::getLaunchFlag` call. See `Mod::getLaunchArgument`
          * for details about mod-specific launch arguments.
          * @param name The argument name
          */
-        bool getLaunchBool(std::string_view const name) const;
+        bool getLaunchFlag(std::string_view const name) const;
```
altalk23, line 181: same here

loader/src/loader/LoaderImpl.cpp
```diff
@@ -559,6 +559,12 @@ void Loader::Impl::refreshModGraph() {
 
     m_problems.clear();
 
+    if (Loader::get()->getLaunchFlag("safe-mode")) {
+        log::debug("Launched in Safe Mode. Skipping mod loading...");
+        m_loadingState = LoadingState::Done;
```
altalk23, line 564: does this actually work sounds hacky

altalk23: スクリーンショット 2024-01-26 1 14 39

loader/include/Geode/loader/Loader.hpp
acikek: We've discussed this a bit in Discord -- presumably it shouldn't affect anyone

loader/src/loader/LoaderImpl.cpp
acikek: I've tested it to the best of my ability and it works fine

loader/include/Geode/loader/Mod.hpp
matcool: this is fine because no mod uses this yet

loader/src/loader/LoaderImpl.cpp
cgytrus: after taking a closer look at this code, looks fine to me, but i pushed a commit that slightly changed the message and added a missing log::popNest() call

PR #461, opened 2024-01-26
Add methods for requesting and querying permissions

dankmeme01:
Adds methods `Loader::getPermissionStatus` and `Loader::requestPermission`, and implements them on Android. Note that this PR only makes sense in junction with https://github.com/geode-sdk/android-launcher/pull/12 and will not work on other launcher builds.


loader/src/platform/android/LoaderImpl.cpp
```diff
@@ -44,3 +44,49 @@ bool Loader::Impl::supportsLaunchArguments() const {
...
+    if (JniHelper::getStaticMethodInfo(info, "com/geode/launcher/utils/GeodeUtils", "getPermissionStatus", "(Ljava/lang/String;)Z")) {
+        jstring permString = info.env->NewStringUTF(std::string(name).c_str());
+        jboolean result = info.env->CallStaticBooleanMethod(info.classID, info.methodID, permString);
+        info.env->DeleteLocalRef(info.classID);
+        info.env->DeleteLocalRef(permString);
+
+        return result == JNI_TRUE;
+    }
+
+    return false;
```
qimiko, line 59: when you're using new methods from the jni, keep in mind that this isn't enough to prevent a crash if the method isn't found. you can see the clear exception code (which you should add) above in `getGameVersion` (the function is also provided in [utils](https://github.com/geode-sdk/geode/blob/a8ab0e4ba69704a1d9859f265a646cefa4cd3aae/loader/src/platform/android/util.cpp#L46), but it's anonymous so you'll have to copy the contents or make a new function here)

loader/src/platform/android/LoaderImpl.cpp
```diff
@@ -44,3 +44,49 @@ bool Loader::Impl::supportsLaunchArguments() const {
...
+    if (JniHelper::getStaticMethodInfo(info, "com/geode/launcher/utils/GeodeUtils", "requestPermission", "(Ljava/lang/String;)Z")) {
+        jstring permString = info.env->NewStringUTF(std::string(name).c_str());
+        jboolean result = info.env->CallStaticBooleanMethod(info.classID, info.methodID, permString);
+        info.env->DeleteLocalRef(info.classID);
+        info.env->DeleteLocalRef(permString);
+
+        if (result == JNI_TRUE) {
+            return Ok();
+        }
+    }
```
qimiko, line 89: same thing here, add exception clearing code :pray:

qimiko (changes requested):

qimiko (commented):
jni stuff looks good :+1:
will probably have to get someone else to approve the method placement though

qimiko (approved):

dankmeme01:
as requested, all the new methods have now been moved to `geode::utils::permission` instead of being in `Loader`.

Fleeym:
I suggest creating an enum with the most used permissions (media permissions in android 13 mostly)

dankmeme01:
> I suggest creating an enum with the most used permissions (media permissions in android 13 mostly)

done :3

Fleeym:
very cool
what about other platforms? when compiling on windows and mac those functions will be undefined

dankmeme01:
> very cool what about other platforms? when compiling on windows and mac those functions will be undefined

they are already defined; `getPermissionStatus()` will always return `true` and `requestPermission` will be a no-op on all platforms that don't support permissions (currently all of them except Android.)

Fleeym:
oh im blind then, all cool

matcool (approved):

PR #467, opened 2024-01-28
update fix for GLFW error handler to 2.204

sse2:
instead of copying function over, patch call ds:MessageBoxW to call ds:MessageBoxA
this is better because now there's 2 different functions for GLFW error handling(???) and they have different calling conventions

before https://i.imgur.com/HCp7d1v.png
after https://i.imgur.com/zzPELWK.png

### DIFF
--- a/loader/src/hooks/MessageBoxFix.cpp
+++ b/loader/src/hooks/MessageBoxFix.cpp
@@ -29,10 +29,24 @@ static void __cdecl fixedErrorHandler(int code, char const* description) {
 }
 
 $execute {
-    // TODO: i'm pretty sure this patch only works on 2.113?
-    //(void)Mod::get()->patch(
-    //    reinterpret_cast<void*>(geode::base::getCocos() + 0x19feec), toByteArray(&fixedErrorHandler)
-    //);
+    // updated for 2.204
+    // check xrefs to "GLFWError #%d Happen, %s\n", now there's two functions with the same exact behaviour, one is a member function though...
+	// call ds:MessageBoxW 
+    // patch MessageBoxW to MessageBoxA
+    // geode::base::getCocos() + 0x122600 = MessageBoxA in .idata
+    // geode::base::getCocos() + 0x1225DC = MessageBoxW in .idata
+    const uint32_t importedMessageBoxA = (geode::base::getCocos() + 0x122600);
+    
+    ByteVector p = {
+        (importedMessageBoxA) & 0xff,
+        (importedMessageBoxA >> 8) & 0xff,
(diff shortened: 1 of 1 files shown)


loader/src/hooks/MessageBoxFix.cpp
```diff
@@ -28,11 +28,25 @@ static void __cdecl fixedErrorHandler(int code, char const* description) {
...
-    //    reinterpret_cast<void*>(geode::base::getCocos() + 0x19feec), toByteArray(&fixedErrorHandler)
-    //);
+$execute {
+    // updated for 2.204
+    // check xrefs to "GLFWError #%d Happen, %s\n", now there's two functions with the same exact behaviour, one is a member function though...
+	// call ds:MessageBoxW 
+    // patch MessageBoxW to MessageBoxA
+    // geode::base::getCocos() + 0x122600 = MessageBoxA in .idata
+    // geode::base::getCocos() + 0x1225DC = MessageBoxW in .idata
+    const uint32_t importedMessageBoxA = (geode::base::getCocos() + 0x122600);
```
matcool, line 38: why the ()

loader/src/hooks/MessageBoxFix.cpp
```diff
@@ -28,11 +28,25 @@ static void __cdecl fixedErrorHandler(int code, char const* description) {
...
+    
+    ByteVector p = {
+        (importedMessageBoxA) & 0xff,
+        (importedMessageBoxA >> 8) & 0xff,
+        (importedMessageBoxA >> 16) & 0xff,
+        (importedMessageBoxA >> 24) & 0xff};
+
+    (void)Mod::get()->patch(reinterpret_cast<void*>(geode::base::getCocos() + 0xC75F9), p
+    );
```
matcool, line 47: why the ); on another line..

matcool (changes requested):

matcool:
if possible this should check for if the loader is in forwardCompatMode, except i think $execute is too early for that?

loader/src/hooks/MessageBoxFix.cpp
sse2: forgot to remove because i casted before, fixing rn

loader/src/hooks/MessageBoxFix.cpp
sse2: visual studio stinks im fixing rn

sse2:
Added check for forwardcompatmode, works fine, checked by manually changing the timestamp of geometrydash.exe

PR #468, opened 2024-01-29
Refactor console handling

cgytrus:
generally works better with debuggers and such on windows now
- console no longer opened/closed in run-time (only opened on startup if enabled)
- log should be fully thread safe now (?)
- windows: the console can now attach to the parent console if ran from cmd or a debugger or whatever
- windows: stdin/stdout are no longer redirected to console, logger uses WriteFile instead
- windows: stdout/stderr are now redirected to logger

### DIFF
--- a/loader/include/Geode/loader/Log.hpp
+++ b/loader/include/Geode/loader/Log.hpp
@@ -109,7 +109,8 @@ namespace geode {
         private:
             class Impl;
             std::shared_ptr<Nest::Impl> m_impl;
-            friend class Logger;
+            friend GEODE_DLL std::shared_ptr<Nest> saveNest();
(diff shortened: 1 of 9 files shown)


matcool:
code looks fine, though still breaks current geode extension "Launch Geometry Dash". i dont even know how but it just doesnt open the game. maybe we should just update it instead of trying to fix it here

matcool (approved):

HJfod:
> code looks fine, though still breaks current geode extension "Launch Geometry Dash". i dont even know how but it just doesnt open the game. maybe we should just update it instead of trying to fix it here

Oh what the hell, I'll look into that tomorrow

PR #476, opened 2024-01-31
Anchor layout

HJfod:
The main interest of this PR is the new `AnchorLayout`, a layout that allows positioning nodes to anchored positions with offsets in their parent. This is a very useful tool alongside the existing `RowLayout` and `ColumnLayout`, as it can represent most of the layouts those two don't cover, such as popups. This means that a popup properly implementing `AnchorLayout` can be re-content-sized without needing to manually move anything around.

As a proof-of-execution, this PR implements the new `AnchorLayout` model for `ModInfoPopup` - however, it does not implement it for any other UIs, as this PR considers those better fit for a more comprehensive reworking of the internal UIs. You can see the effects of `AnchorLayout` by resizing the `m_mainLayer` of a `ModInfoPopup` with DevTools, and seeing how the different parts of the popup react and adjust themselves to the changes.

image

Since `AnchorLayout`s are more about the child's positioning and not about the parent's layout, there has been introduced a new function to `CCNode` called `addChildAtPosition` for syntactic sugar in setting anchored positions. This function can also be used to conveniently position the node even if the parent doesn't use `AnchorLayout` - `AnchorLayout` provides dynamic position updates on resizing.

This PR does not introduce any breaking changes since `Popup` is a templated class, except for a visual breaking change with `MDTextArea` no longer having `ignoreAnchorPointForPosition` true. This will break approximately 0 mods as all the uses of this node are inside the loader and those places have been fixed in this PR.

## Changes

 * Add `AnchorLayout` for anchored positioning
 * Add `CCNode::addChildAtPosition` for more easily positioning nodes
 * Add `CopySizeLayout` for automatically mirroring the content size of a node to specified targets
 * Deprecate `Popup::init` and add `Popup::initAnchored` for initializing a popup with `AnchorLayout` as well as tying the size of `m_buttonMenu` and `m_mainLayer` to the size of the background
 * Split layout implementations out into their own files rather than dumping everything in `Layout.cpp`
 * Make `MDTextArea` no longer ignore anchor point for position
 * Rework the internal UIs of `ModInfoPopup` and its derivatives to fix a few minor bugs and use the new layouts


altalk23:
holy boobs

PR #479, opened 2024-02-01
implement getGameVersion on mac

dankmeme01:
i think it even works


cgytrus:
does this return the actual gd version? like, in the `2.113`, `2.200`, etc format (this has to be consistent with all platforms)

qimiko:
can confirm it works and is also 2.200 (good ending)

PR #481, opened 2024-02-02
Use Breakpad for crash reports on Android 

qimiko:
as an alternative to the current (very broken) crash handler. ideally the original crash handler would be fixed, but this is a workable solution for now. just adds a lot of complexity to the act of sharing crash dumps

also: `-DGEODE_USE_BREAKPAD=OFF` will disable it for a build, adding `--geode:disable-crash-handler` will disable it for a session

## TODOs

- [x] find some way to generate the necessary breakpad tools (`dump_syms` + `minidump_stackwalk`)
- [x] add symbol exporting to github action builds
- [x] maybe strip the main geode binaries too

## Future Changes

- would be good to have some sort of symbol server to make the process of resolving symbol dumps a little easier
- breakpad stores dumps with a very strange naming system, which may be a little hard to find. imo the crash popup should redirect people to export the last crash

### DIFF
--- a/loader/CMakeLists.txt
+++ b/loader/CMakeLists.txt
@@ -126,7 +126,7 @@ if (WIN32)
 endif()
 
 
-if (ANDROID AND GEODE_ANDROID_USE_BREAKPAD)
+if (ANDROID AND GEODE_USE_BREAKPAD)
 	CPMAddPackage(
 		NAME "Breakpad"
 		VERSION "2023.06.01"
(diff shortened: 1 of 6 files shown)


loader/CMakeLists.txt
```diff
@@ -125,6 +125,61 @@ if (WIN32)
...
+
+		FetchContent_Declare(lss
+				GIT_REPOSITORY https://chromium.googlesource.com/linux-syscall-support.git
+				GIT_TAG v2022.10.12
+				SOURCE_DIR ${Breakpad_BINARY_DIR}/src/third_party/lss
+		)
+
+		FetchContent_MakeAvailable(lss)
+
+			add_library(Breakpad STATIC
```
matcool, line 147: y the random indentation increase

qimiko: lol oops 😭 must've been a bad copy paste

CAIify3623:
hi guys, do you know why the qqlmod crashes geode? please fix that, i want to use it again

PR #484, opened 2024-02-03
Run codegen on configure

cgytrus:
depends on https://github.com/geode-sdk/bindings/pull/175


CMakeLists.txt
```diff
@@ -143,120 +142,39 @@ if (DEFINED ENV{GEODE_BINDINGS_REPO_PATH})
 	set(temp $ENV{GEODE_BINDINGS_REPO_PATH})
 	# this is so stupid i hate windows paths
 	string(REPLACE "\\" "/" GEODE_BINDINGS_REPO_PATH ${temp})
+	add_subdirectory(${GEODE_BINDINGS_REPO_PATH} bindings)
```
matcool, line 145: um wouldnt you want this inside the build folder?

matcool: ok nvm mb
>  The binary_dir specifies the directory in which to place the output files. If it is a relative path, it will be evaluated with respect to the current output directory, but it may also be an absolute path.
https://cmake.org/cmake/help/latest/command/add_subdirectory.html

matcool (approved):

CMakeLists.txt
cgytrus: so that it doesnt accidentally mess with anything in the main build folder

PR #491, opened 2024-02-05
Better patching

HJfod:
Deprecates the existing `Mod::patch` API and adds platform-specific patching APIs. Reasoning is that since patching is inherently unportable, this reminds modders better that they need to also work out the correct patch for other platforms.

### Unresolved questions
 - [x] Implement platform-specific APIs
 - [x] Add warning to `Mod::patch`
 - [ ] Should the functions issue a warning if you try to call them on another platform?
 - [ ] Should `Mod::patch` be removed completely?


matcool:
this stinks sorry

PR #501, opened 2024-02-09
Fix Wine crashing when launching via terminal

FireMario211:
These lines cause Geode to crash (without any crash logs) when launching Geometry Dash through the terminal. Or `wine GeometryDash.exe`

Removing these lines fixes the issue, though @cgytrus requested that @dankmeme01 & @matcool (also themselves) test it.

Let me know if this causes any issues.

### DIFF
--- a/loader/src/platform/windows/console.cpp
+++ b/loader/src/platform/windows/console.cpp
@@ -113,6 +113,19 @@ void console::setup() {
             if (count != 0) {
                 path = std::string(buf, count - 1);
             }
+
+            // count == 0 => not a console and not a file, assume it's closed
+            // wine does something weird with /dev/null? not sure tbh but it's definitely up to no good
+            if ((count == 0 || path.ends_with("\\dev\\null")) && Mod::get()->getSettingValue<bool>("show-platform-console")) {
+                s_outHandle = nullptr;
+                CloseHandle(GetStdHandle(STD_OUTPUT_HANDLE));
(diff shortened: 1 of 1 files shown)


cgytrus:
made it a draft for now, i'll test when i have time

dankmeme01:
caused the platform console to not appear on proton GE for me.

FireMario211:
> caused the platform console to not appear on proton GE for me.

I added the if condition back, but only did it if platform console is enabled. It should work, though platform console wont if ran in the terminal.

Though it should send to the output anyways when ran with the terminal.

Cvolton:
> caused the platform console to not appear on proton GE for me.

platform console doesnt appear for me in proton even on current stable

FireMario211:
> > caused the platform console to not appear on proton GE for me.
> 
> platform console doesnt appear for me in proton even on current stable

dont think it's related to this PR. While I didn't try Proton GE, platform console works fine for me.

matcool:
isnt this kind of a jank fix? if a wine user turns on platform console then theyll get the crash the same way..

FireMario211:
> isnt this kind of a jank fix? if a wine user turns on platform console then theyll get the crash the same way..

No, it'll only show platform console **when the game isn't launched via terminal.**
It won't crash, I tried it myself with the "Show Platform Console" setting on.

Launching with Terminal
image

Launching without Terminal
image

PR #565, opened 2024-02-25
fix pch being treated as c files

iAndyHD3:
This pr fixes cmake treating precompiled headers as C files. CMake generates .c and .h files instead of .cxx and .hxx which causes a compiler error, since precompiled headers get compiled in C mode. So far it seems like I'm the only one that has experienced this issue. The proposed solution is a common practice in cmake, as seen by the [1.1k github search results](https://github.com/search?q=%22target_precompile_headers%22+%3C%24%3CCOMPILE_LANGUAGE%3ACXX%3E+language%3ACMAKE&type=code)


altalk23:
someone touched cmake :pray:

PR #589, opened 2024-03-05
Fix cocos TTF font caching issues on Windows

glistermelon:
## Problem

For whatever reason, in my experience on all of my Windows systems with very basic testing using `CCLabelTTF`, cocos does not properly remove user provided (via TTF file) font resources from the font table. After closing the app, any user-provided font files are impossible to delete without killing the Windows Font Cache Service. This makes it impossible to delete the respective mod's active save directory (or "temp dir"), which causes a whole host of other problems (uninstalling the mod, updating the mod, uninstalling Geode, etc).

The implementation causing the problems in question should be [here](https://github.com/cocos2d/cocos2d-x/blob/cocos2d-x-2.2.3/cocos2dx/platform/win32/CCImage.cpp). I don't particularly see anything wrong with it but I'm definitely having problems.

## Reproducing the Issue

The minimal code that caused this problem for me was the following. **I hope I'm not misusing `CCLabelTTF` here.**
```cpp
class $modify(MenuLayer) {
    bool init() {
        if (!MenuLayer::init()) return false;
        auto label = CCLabelTTF::create("Hello World", "Font.ttf"_spr, 50.0f);
        this->addChild(label);
    }
}
```
To reproduce with the above code:
* Build the mod
* Run the game with the mod enabled (if the text that appears isn't of the font you provided, the issue won't be reproduced)
* Close the game
* Rebuild the mod after changing anything in it (so the hash changes) (I usually change the font size specified in `CCLabelTTF::create`)
* Run the game again
* And then on my systems, I get an error as Geode is unable to delete the "temp dir."

## TODO

* Fix it (ignore the commits they don't work lol)
* More testing (I know I've never actually reached one of the three patched calls in my testing so far so...)
* Verify that the patched functions can still be hooked and called from hooks with the patches intact


glistermelon:
I imagine it's also possible to just completely rewrite the function to exhibit preferable behavior without any trampoline but it wasn't the first thing I thought of.

glistermelon:
Leaving as a draft until I fix a few problems.

glistermelon:
Closing this because the commit count is uncomfortable but I'll most likely open another one with fewer commits

TheBearodactyl:
> Closing this because the commit count is uncomfortable but I'll most likely open another one with fewer commits

just rebase and squash?

PR #595, opened 2024-03-06
Implemented Windows Font Cache fix (again)

glistermelon:
I opened an earlier PR but it was a train-wreck so I'm back again with one that shouldn't be hot garbage (maybe just normal garbage). Anyway most of this is copied over:

## Problem

For whatever reason, in my experience on all of my Windows systems with very basic testing using CCLabelTTF, cocos does not properly remove user provided (via TTF file) font resources from the font table. After closing the app, any user-provided font files are impossible to delete without killing the Windows Font Cache Service. This makes it impossible to delete the respective mod's active save directory (or "temp dir"), which causes a whole host of other problems (uninstalling the mod, updating the mod, uninstalling Geode, etc).

The implementation causing the problems in question should be [here](https://github.com/cocos2d/cocos2d-x/blob/cocos2d-x-2.2.3/cocos2dx/platform/win32/CCImage.cpp). I don't particularly see anything wrong with it but I'm definitely having problems.
Fix

All the fix does is hook [BitmapDC::setFont](https://github.com/cocos2d/cocos2d-x/blob/5a25fe75cb8b26b61b14b070e757ec3b17ff7791/cocos2dx/platform/win32/CCImage.cpp#L96) so that it immediately releases any user-provided (in the active save directory) fonts. I haven't had any performance issues with this, but if issues are a concern, I could easily implement an alternate implementation that keeps track of added user fonts and releases them whenever the app closes or crashes, but any unexpected behavior beyond what Geode's crash handler can handle would cause problems.

Regardless, assuming I'm not misusing them, I think a fix for this issue ought to be included in the loader or some other part of the framework because having to include Windows-specific fixes on a mod-by-mod basis any time TTF labels are used doesn't line up with Geode's cross-platform goals.

## Reproducing the Issue

The minimal code that caused this problem for me was the following. I hope I'm not misusing CCLabelTTF here.

class $modify(MenuLayer) {
    bool init() {
        if (!MenuLayer::init()) return false;
        auto label = CCLabelTTF::create("Hello World", "Font.ttf"_spr, 50.0f);
        this->addChild(label);
    }
}

To reproduce with the above code:

* Build the mod
* Run the game with the mod enabled (if the text that appears isn't of the font you provided, the issue won't be reproduced)
* Close the game
* Rebuild the mod
* Run the game again
* And then for me, I get an error as Geode is unable to delete the "temp dir."

## Fix Attempt

Patch calls to `AddFontResourceW` and `RemoveFontResourceW`, made in the `BitmapDC::~BitmapDC` and `BitmapDC::setFont` functions, to call their respective extended versions with the `FR_PRIVATE` flag set, so that Windows releases the font resources whenever the app terminates.

## TODO

* This "fix" has worked for me like 99% of the time but I've had one single instance where it failed (notably the file was marked as open by "System" instead of the Font Cache Service) which I've been unable to reproduce. So if anybody knows anything about that...
* More testing; need to experiment with switching between multiple fonts at runtime to ensure every patch is hit

### DIFF
--- a/loader/src/hooks/WindowsFontCacheFix.cpp
+++ b/loader/src/hooks/WindowsFontCacheFix.cpp
@@ -1,0 +1,46 @@
+#include <Geode/Geode.hpp>
+
+#ifdef GEODE_IS_WINDOWS
+
+using namespace geode::prelude;
+
+// https://github.com/cocos2d/cocos2d-x/blob/5a25fe75cb8b26b61b14b070e757ec3b17ff7791/cocos2dx/platform/win32/CCImage.cpp#L96
+// stop setFont from caching fonts on Windows
+// Windows Font Cache Manager holds on to them after the app has closed
+// Fonts are supposed to be released in the BitmapDC::~BitmapDC() but this doesn't seem to work consistently?
+// Looks like there's a shared instance of BitmapDC, so presumably it's destructed when the app closes?
+
+int __stdcall AddFontResourceWHook(volatile LPCWSTR p0) {
+    return AddFontResourceExW(p0, FR_PRIVATE, 0);
+}
+
+int __stdcall RemoveFontResourceWHook(volatile LPCWSTR p0) {
+    return RemoveFontResourceExW(p0, FR_PRIVATE, 0);
+}
+
+/*
+ * addr is relative to cocos base
+ * patches x86 CALL at addr to call function at newCall
+ */
+static void patchCall(uintptr_t addr, uintptr_t newCall) {
+    ByteVector patch = { 0xE8 }; // CALL near & relative
+    addr += (uintptr_t)geode::base::getCocos();
+    uintptr_t callAddr = newCall - (addr + 5);
+    for (auto i = 0; i < sizeof(int); ++i)
+        patch.push_back(callAddr >> (8 * i));
+    patch.push_back(0x90); // every overwritten instruction happens to be 6 bytes wide
+    (void)Mod::get()->patch(reinterpret_cast<void*>(addr), patch);
+}
+
+$execute {
+
+    // BitmapDC::~BitmapDC
+    patchCall(0xC9A56, (uintptr_t)&RemoveFontResourceWHook);
+
+    // BitmapDC::setFont
+    patchCall(0xCB5BC, (uintptr_t)&RemoveFontResourceWHook);
+    patchCall(0xCB642, (uintptr_t)&AddFontResourceWHook);
+
+};
+
+#endif


glistermelon:
> can u credit me at least lol

Sorry but what for? I don't recall interacting with you at all?

cloudskies13:
because i made a pr adds that patchcall thing

glistermelon:
> because i made a pr adds that patchcall thing

Oh; In full honesty I wrote all of this myself without even noticing that you had forked it. The code is really similar though, so I can try to add you as an author to the initial commit (if git allows) since I basically did write the same thing though.

cloudskies13:
atleast credit me for the idea
and you added that thing 2 days ago, same day i made a pr

glistermelon:
> atleast credit me for the idea and you added that thing 2 days ago, same day i made a pr

I'm sorry but I'm being completely honest when I say that I never saw your PR, which is my bad. I do acknowledge that you wrote that first and I'm willing to credit you for that. I'm willing to add a note if that's really all you want (and in that case are you looking for a note in the code or a note in the PR description). But, I can't just bluntly say that you wrote X and I didn't write X, while I can note what actually happened as we've been discussing.

The main reason I jumped to just adding authorship (which I'm still fine with doing) is that the optimal solution here would just be to substitute my function with yours, possibly modifying the it a little bit (they are a little different), but I don't want to muddy up the commit history to replace working code with nearly identical working code solely for the purpose of attributing credit unless one of the geode maintainers prefers that. I think co-authorship of the initial commit (I just took all the commits from the old repo and mashed them into one) does a good job of crediting your work and dispelling all the weird nuance here without making this PR unnecessarily complex for reviewers. Adding notes to the code or to the PR description is more difficult because in those cases I can't really disregard the nuance easily.

glistermelon:
Also, I do want to say that I'm being kinda generous here because my implementation looks like it was actually committed first. Correct me if I'm wrong but:
* [Mine at 6:22 AM CST](https://github.com/geode-sdk/geode/pull/589/commits/5652f5ebb961c02ed25b8b65e2b02b72a7c8f612)
* [Yours at 8:20 AM CST](https://github.com/cloudskies13/geode/commit/3236a16102fde055d74d1014d6610db40445edd9)

cloudskies13:
oh youre right ig? srry

glistermelon:
I'm going to mark this as ready for review because I've done as much testing on my end as possible, though I would highly encourage someone else to test it as well.

cloudskies13:
okaay

altalk23:
not the prettiest but i suppose it's fine

PR #625, opened 2024-03-14
crashlog: add method names from bindings

Prevter:
Adds a basic system to read a "GeometryDash.bro" file (currently has to be placed manually into `Geometry Dash/geode` directory) and get function names from it to include them in the stacktrace.
This change makes crashlog look like this:
image

Unknown methods are also easier to find, because it prints the address where the method starts, instead of only the instruction address.

Some comments/questions:
- What would be the best way to implement auto download of the bindings file (at which stage to do it, and where to actually save it)?
- Maybe provide a toggleable option inside mod configuration to redownload bindings file automatically on every launch/crash?

Waiting for your feedback :)

### DIFF
--- a/loader/src/platform/windows/crashlog.cpp
+++ b/loader/src/platform/windows/crashlog.cpp
@@ -90,6 +90,67 @@ static Mod* modFromAddress(PVOID exceptionAddress) {
     return nullptr;
 }
 
+static uintptr_t findMethodStart(void const* address, size_t maxSearch = 0x1000) {
+    // Backtrack until we hit a 0xCC (int3) instruction
+    uintptr_t start = reinterpret_cast<uintptr_t>(address);
+    while (start > reinterpret_cast<uintptr_t>(address) - maxSearch) {
+        if (*reinterpret_cast<uint8_t*>(start) == 0xCC) {
+            return start + 1;
+        }
+        start--;
+    }
+    return reinterpret_cast<uintptr_t>(address);
+}
+
+static std::string formatMethodName(std::string const& className, std::string const& line) {
+    // strip everything after first '=' sign to make sure we don't mess with text in the comments
+    auto eq = line.find('=');
+    if (eq != std::string::npos) {
+        return formatMethodName(className, line.substr(0, eq));
+    }
+
+    // Get only the method name (last word before '(')
+    auto end = line.find('(');
+    if (end == std::string::npos) {
+        return line; // this should never happen, but just in case
+    }
+    auto start = line.rfind(' ', end);
+    if (start == std::string::npos) {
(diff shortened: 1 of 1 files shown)


poweredbypie:
This is a pretty cool addition, very impressive!
Although, I'm not sure if this is a good approach to implementing it: This requires us to ship the Broma binding with the loader, along with being kind of hacky about parsing it (we already have the broma parser library included in the source tree!)
I think the appropriate way to go about implementing this would be for the codegen library to emit some sort of address-to-symbol table so you can just lookup an address directly instead. This would mean no extra code units shipped (it's just included in the loader DLL) and automatically handles platform issues (just use the parsed broma to emit only the correct bindings for the respective platform).

Prevter:
Thanks for your feedback, I will try to implement your proposals.

> I'm not sure if this is a good approach to implementing it
I agree, because I'm not familiar with internal structure of the project. That's why I wanted to hear some thoughts.

> we already have the broma parser library included in the source tree
Didn't know about that, will check it. 😆

> I think the appropriate way to go about implementing this would be for the codegen library to emit some sort of address-to-symbol table so you can just lookup an address directly instead
Yeah, this actually sounds like a more clean solution. The only downside is probably that bindings file would only be updated alongside loader updates, but I guess that it doesn't need to always have most up-to-date bindings to begin with. Also I don't think it would apply to anything other than Windows because of how bindings work on other platforms (maybe it might work on macOS, but I can't be sure).

altalk23:
codegen generates a CodegenData.txt file, i think this can be embbeded instead

Prevter:
Figured out how to embed CodegenData.txt into the binary (codegen for codegen :0), so not it works without having to store the broma file locally.  
CMake codegen function haven't increased configuration time at all for me, but crashlog.cpp may build for a few seconds longer now (i think this change is worth it in the long run).

prevter:
Since everything was remade from scratch, here's new list of changes:
- Bindings (for the exact platform) are downloaded once per day and stored in geode save dir
- Instead of doing hacky function boundary search, do a quick binary search on the bindings
- Ported the functionality to macOS (Arm/Intel) and iOS
- Modernized crashlogs to use StringBuffer instead of std::stringstream (where applicable)
- On macOS, stripped the game directory path, to avoid long lines
- Holding shift before pressing "Restart game" button in Windows crashlog window, prompts to save the game
- Handles libcocos2d.dll having missing symbols on Wine sometimes

One remaining problem is how should the bindings files be hosted? Right now it's hardcoded to my [bindings-meta](https://prevter.github.io/bindings-meta) repo, which has them for every game version/platform and update automatically.

altalk23:
lgtm

PR #656, opened 2024-04-04
getChildOfType negative indices: size_t is unsigned

nytelytee:
`i` is always >= 0 because it is unsigned. This causes crashes if no children are found before reaching 0.


altalk23:
aah the most classic bug

PR #666, opened 2024-04-08
FMODQueuedMusic is not an enum

0x5abe:


altalk23:
can you make the corresponding bindings pr as well, since merging this as is will cause errors

0x5abe:
> can you make the corresponding bindings pr as well, since merging this as is will cause errors (adding the structs to Extras.bro in all of the versions)

Sure thing but just as a warning there is a bunch of other stuff as well, not only the new structs for types that were enums before. That's why I was waiting a bit to make the bindings PR

PR #706, opened 2024-04-19
GJDifficultyName and GJFeatureState

hiimjustin000:


altalk23:
nice GJFeatureState
ill merge when you fix

hiimjustin000:
Alk I fixed it can you merge now

loader/include/Geode/Enums.hpp
```diff
@@ -257,8 +257,17 @@ enum class StatKey {};
 enum class TextStyleType {};
 enum class InputValueType {};
 enum class GJInputStyle {};
-enum class GJDifficultyName {};
-enum class GJFeatureState {};
+enum class GJDifficultyName {
```
SpaghettDev, line 260: i think this is wrong, logging it to the console in GJDifficultySprite::create results in 0 and 1 appearing on both normal and demon difficulties

hiimjustin000: No it is not.

SpaghettDev: non-demon:
image

demon:
image

it does appear that 0 is passed to GJDifficultySprite::create when in LevelBrowserLayer/LevelListLayer, but in LevelInfoLayer 1 is used. While GJDifficultySprite::updateDifficultyFrame is always passed 1?

hiimjustin000: Let me explain my stance:
If `GJDifficultyName` is 1, the demon difficulty will be shown in text if it is a demon, otherwise, it will revert to default.
It is condensed if `GJDifficultyName` is 0.
`updateDificultyFrame` is only used in `LevelInfoLayer`.

SpaghettDev: yeah so my last point still stands, a more appropriate name would be DefaultText and ShortText imo

PR #791, opened 2024-05-26
Update updater.cpp - downloadLatestLoaderResources

staFF6773:
# Details of the Improvements

1- More Informative Logging: Added an error message when the "resources.zip" file is not found in the release assets.

2 - Use of Temporary Variables: Used temporary variables for the asset name (assetName) and the download URL (downloadUrl) to improve clarity.

These changes are small but make the code easier to read and correct without altering its operation.

### DIFF
--- a/loader/src/loader/updater.cpp
+++ b/loader/src/loader/updater.cpp
@@ -82,33 +82,37 @@ void updater::fetchLatestGithubRelease(
 }
 
 void updater::downloadLatestLoaderResources() {
-    log::debug("Downloading latest resources", Loader::get()->getVersion().toString());
+    // Log the start of the download process
+    log::debug("Downloading latest resources for version: ", Loader::get()->getVersion().toString());
+
+    // Fetch the latest GitHub release
     fetchLatestGithubRelease(
+        // Lambda function to process the JSON response
         [](matjson::Value const& raw) {
             auto json = raw;
             JsonChecker checker(json);
             auto root = checker.root("[]").obj();
 
-            // find release asset
+            // Find the release asset named "resources.zip"
             for (auto asset : root.needs("assets").iterate()) {
                 auto obj = asset.obj();
-                if (obj.needs("name").template get<std::string>() == "resources.zip") {
-                    updater::tryDownloadLoaderResources(
-                        obj.needs("browser_download_url").template get<std::string>(),
-                        false
-                    );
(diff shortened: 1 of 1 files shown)


loader/src/loader/updater.cpp
```diff
@@ -82,33 +82,37 @@ void updater::fetchLatestGithubRelease(
...
+            log::error("Unable to find 'resources.zip' in the latest GitHub release");
+            ResourceDownloadEvent(UpdateFailed("Unable to find 'resources.zip' in the latest GitHub release")).post();
         },
+        // Lambda function to handle errors during the fetching process
         [](std::string const& info) {
-            ResourceDownloadEvent(
-                UpdateFailed("Unable to download resources: " + info)
-            ).post();
+            std::string errorMsg = "Unable to download resources: " + info;
+            log::error(errorMsg);
```
matcool, line 114: thats not how you use our logging

loader/src/loader/updater.cpp
```diff
@@ -82,33 +82,37 @@ void updater::fetchLatestGithubRelease(
 }
 
 void updater::downloadLatestLoaderResources() {
-    log::debug("Downloading latest resources", Loader::get()->getVersion().toString());
+    // Log the start of the download process
+    log::debug("Downloading latest resources for version: ", Loader::get()->getVersion().toString());
+
+    // Fetch the latest GitHub release
     fetchLatestGithubRelease(
+        // Lambda function to process the JSON response
```
matcool, line 90: are these nothing comments really necessary?

loader/src/loader/updater.cpp
```diff
@@ -82,33 +82,37 @@ void updater::fetchLatestGithubRelease(
...
+            // Find the release asset named "resources.zip"
             for (auto asset : root.needs("assets").iterate()) {
                 auto obj = asset.obj();
-                if (obj.needs("name").template get<std::string>() == "resources.zip") {
-                    updater::tryDownloadLoaderResources(
-                        obj.needs("browser_download_url").template get<std::string>(),
-                        false
-                    );
+                std::string assetName = obj.needs("name").template get<std::string>();
+                if (assetName == "resources.zip") {
```
matcool, line 100: were these changes really necessary? no point in storing the string in a local variable, and also this might be implicitly creating a copy

matcool (changes requested):

PR #797, opened 2024-05-27
Make getGameVersion public

Prevter:
Can be useful to some mods that have `"gd": "*"` in their `mod.json`


altalk23:
🔥 blazingly 🚀 fast 🔥

PR #830, opened 2024-06-02
feat(win): add 2.206 timestamp and update gd in cmake

RoootTheFox:
disables forward compat


RoootTheFox:
mat impl'd the same thing in a7bf4ef5ae4cb0c4b8c896bcd9151a1680611bd6

PR #838, opened 2024-06-03
update cmake shit

TheBearodactyl:


matcool:
ok two things:
* i think it would be better off just renaming the current win32 to win64, no point in having both given the game only supports 64 bit
* WIN64 is not a real cmake variable, WIN32 covers both arches

matcool:
another thing, renaming to Win64 will most definitely break codegen, since that excepts Win32, so its something we have to deal with there as well :-)

TheBearodactyl:
> ok two things:
> 
> * i think it would be better off just renaming the current win32 to win64, no point in having both given the game only supports 64 bit
> * WIN64 is not a real cmake variable, WIN32 covers both arches

what if people still want to build for earlier versions?

TheBearodactyl:
> another thing, renaming to Win64 will most definitely break codegen, since that expects Win32, so its something we have to deal with there as well :-)

fair point

matcool:
> > ok two things:
> > 
> > * i think it would be better off just renaming the current win32 to win64, no point in having both given the game only supports 64 bit
> > * WIN64 is not a real cmake variable, WIN32 covers both arches
> 
> what if people still want to build for earlier versions?

the geode repo itself is only meant for the latest gd version, things like cocos headers and the stub libraries would have to be changed anyways, which is much more work. they can go back in git history and copy the relevant 32 bit code if they want

matcool:
closing because we already renamed the platform to win64

PR #858, opened 2024-06-06
Dtor fix

SMJSGaming:
$modify doesn't like private dtors
image


altalk23:
does this actually compile on windows? since windows has access modifier in mangle

Fleeym:
seems to compile

SMJSGaming:
it compiles on windows, but $modify can't compile on windows with CCHttpClient in the old version

PR #868, opened 2024-06-08
Fixed padding and added a content layer to borders

SMJSGaming:


altalk23:
youre lucky were still in alpha

PR #877, opened 2024-06-09
UI Fixes and Revamped Developer List

Alphalaneous:
This PR fixes a ton of minor issues with the UI, including:

- Fixes misaligned Download counts due to non absolute button widths
- Fixes misaligned checkboxes due to content size changing from different sprites
- Replaces the exclamation point on the geode logo with the vanilla one
- Fixes the incorrect mod settings popup size that would get clipped on the top otherwise
- Fixes the oddly short mod problems popup
- Fixes small gap with border on mod settings

This PR also revamps the mod developer list to use a list rather than add a bunch of labels for each developer.

### DIFF
--- a/loader/src/ui/mods/list/ModItem.cpp
+++ b/loader/src/ui/mods/list/ModItem.cpp
@@ -351,6 +351,9 @@ void ModItem::updateState() {
         m_versionLabel->setColor(to3B(ColorProvider::get()->color("mod-list-version-label"_spr)));
     }
 
+    m_viewMenu->updateLayout();
+    m_titleContainer->updateLayout();
+
     // If there were problems, tint the BG red
     if (m_source.asMod() && m_source.asMod()->hasProblems()) {
         m_bg->setColor("mod-list-errors-found"_cc3b);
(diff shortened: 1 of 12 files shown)


loader/src/ui/mods/list/ModItem.cpp
```diff
@@ -379,7 +377,14 @@ void ModItem::updateState() {
             on->setColor(ccGRAY);
             on->setOpacity(105);
         }
+
+        // Make sure toggler is always the same size and position to prevent misalignment
+        m_enableToggle->setContentSize({35, 35});
```
HJfod, line 382: `CCMenuItemToggler::createWithStandardSprites` should just set the content size of the selected sprite to the same as the unselected sprite for a systematic fix of the root issue

HJfod (commented):
i want to see the dev popup before approving

loader/src/ui/mods/list/ModItem.cpp
Alphalaneous: That is what this uses, but it doesn't work, causing the checkbox to shift in the layout when the content size changes

Alphalaneous:
Bindings fail moment

snow111222-glitck2212222312:
it wont connect to the servers, it says "Couldn't resolve host name" can i have some help with this please

Uproxide:
> it wont connect to the servers, it says "Couldn't resolve host name" can i have some help with this please

munch munch crunch crunches your skull

Flobup:
what?

matcool (approved):

PR #888, opened 2024-06-11
Added Norwegian

Ulladay-Hub:
I added Norwegian, I'll add more languages


matcool:
why did you translate the comments as well 😟 ? did you just run this through a translator..

Ulladay-Hub:
no, i translated because i felt a bit bored.

cursefroge:
but why would you translate the comments

PR #894, opened 2024-06-14
Rename `getChildBySpriteName` to `getChildByFilename`

hiimjustin000:
Let me know if there is anything that needs to be changed


matcool:
why the rename?

hiimjustin000:
It gets referred to a filename a lot in cocos.

matcool:
imo old name is fine, and this breaks api and abi so not very nice!

Piecuuu:
You could've not renamed it but clone it and then rename it. The old one would just be marked as deprecated so it doesn't break the api.

PR #917, opened 2024-06-18
Add an aliases features in geode::dirs

YellowCat98:
this pull request adds a few aliases to geode::dirs.

Here's an example:

# Getting the game's directory without geode::dirs::alias:
```c++
fs::path dir = geode::dirs::getGameDir();
```
# Getting the game's directory with geode::dirs::alias:
```c++
fs::path dir = geode::dirs::alias("game://");
// this would return the same thing as geode::dirs::getGameDir();
```

### DIFF
--- a/loader/src/server/DownloadManager.cpp
+++ b/loader/src/server/DownloadManager.cpp
@@ -306,7 +306,7 @@ bool ModDownloadManager::checkAutoConfirm() {
             for (auto inc : confirm->version.metadata.getIncompatibilities()) {
                 // If some mod has an incompatability that is installed,
                 // we need to ask for confirmation
-                if (inc.mod) {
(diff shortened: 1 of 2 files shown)


HJfod:
Why?

matcool:
ignoring the implementation, is it really any shorter/better? you lose the concrete type checked functions (that show up in intellisense) in return of a runtime string check.
```cpp
geode::dirs::getGameDir() / "some/file.txt";
geode::dirs::alias("game://some/file.txt");

geode::dirs::getGeodeSaveDir() / "some/file.txt";
geode::dirs::alias("geode:save://some/file.txt");
```

YellowCat98:
i initially didnt know whether this would be useful or not but since pull requests need to be checked i decided to pr this anyway and let the maintainers decide

PR #935, opened 2024-06-21
Use cross-compilation from Linux for CI Windows builds

cgytrus:


matcool:
reminder to move my repo to the org before this is made official

also no merging before the 22nd

PR #943, opened 2024-06-22
More web options (HTTP version, encoding, transfer body, follow request, range)

SpaghettDev:


matcool:
bro needs to set http version 💀

loader/include/Geode/utils/web.hpp
```diff
@@ -122,12 +134,20 @@ namespace geode::utils::web {
 
         WebRequest& userAgent(std::string_view name);
 
+        WebRequest& encoding(std::string_view encoding);
```
matcool, line 137: encoding of what? the request or the response? and what are valid values?

loader/include/Geode/utils/web.hpp
```diff
@@ -122,12 +134,20 @@ namespace geode::utils::web {
 
         WebRequest& userAgent(std::string_view name);
 
+        WebRequest& encoding(std::string_view encoding);
+
         WebRequest& timeout(std::chrono::seconds time);
 
+        WebRequest& range(std::pair<std::uint64_t, std::uint64_t> byteRange);
```
matcool, line 141: what is this for

loader/include/Geode/utils/web.hpp
```diff
@@ -122,12 +134,20 @@ namespace geode::utils::web {
...
 
+        WebRequest& encoding(std::string_view encoding);
+
         WebRequest& timeout(std::chrono::seconds time);
 
+        WebRequest& range(std::pair<std::uint64_t, std::uint64_t> byteRange);
+
         WebRequest& certVerification(bool enabled);
+        WebRequest& transferBody(bool enabled);
+        WebRequest& followRequest(bool enabled);
```
matcool, line 145: could this be renamed to `followRedirects` or is that a different thing

matcool:
i think some doc comments would help out a bunch of these cuz theyre kinda ambiguous as to what they do

SpaghettDev:
sure

loader/include/Geode/utils/web.hpp
SpaghettDev: [its the encoding of the response](https://curl.se/libcurl/c/CURLOPT_ACCEPT_ENCODING.html), specifically a download response. as for the valid values, it depends on libcurl, but an example of valid values are br, gzip, deflate. these can be passed to encoding as a comma separated string.

loader/include/Geode/utils/web.hpp
SpaghettDev: the range of bytes to download from a response

loader/include/Geode/utils/web.hpp
SpaghettDev: there is no followRedirects lol, it used to be default enabled, this PR made it a default false option. I can rename it to `followRedirects` if you want

matcool: it now defaults to not following redirects? i feel like that is a bad idea, subtle api change :nedfast:

SpaghettDev: ? https://github.com/SpaghettDev/geode/blob/b8509107a971c31a47babe337c8cb83ad29e8083/loader/src/utils/web.cpp#L202

matcool: yes honey, it used to default to true and now, as you said, it defaults to false

SpaghettDev: mb that was a mistake on my part, i meant i made it a default-true option

matcool (approved):

matcool:
@Cvolton any thoughts

Cvolton (approved):
lgtm

PR #950, opened 2024-06-23
Better compression for installer on windows

davidkra230:
This basically just enables solid LMZA compression, makes the compression dictionary size 64mb and, optimizes the data block.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,7 +30,7 @@ jobs:
           id: win
           host_id: win
           extra_flags: ''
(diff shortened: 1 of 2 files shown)


matcool:
why the 64mb dictionary size may i ask?

matcool:
also why not put this in the installer script itself instead of the cli args

matcool:
41.7mb -> 34.5mb pretty nice

davidkra230:
> why the 64mb dictionary size may i ask?

it usually produces better results when compared to the default 8, the tradeoff is ram used when extracting (64 mb extra instead of 8)

cgytrus (approved):

PR #977, opened 2024-06-28
Fixes for TextInput

FlafyDev:
## fixes padding for TextInput
before this PR, when using textalign center there was no padding for text that reached the sides of the input box:
image

and when using textalign left there was padding on the left but on the right it overflowed.
image

this PR fixes both cases to this:
image

## fixes setWidth
before this PR, setWidth doesn't set the width of the TextInput node. So if you set the width to a number smaller than the original width, it'd look like this:
image

and if you set the width to a number larger than the original width:
image

this PR fixes that

### DIFF
--- a/loader/src/ui/nodes/TextInput.cpp
+++ b/loader/src/ui/nodes/TextInput.cpp
@@ -70,7 +70,7 @@ bool TextInput::init(float width, std::string const& placeholder, std::string co
     m_bgSprite->setContentSize({ width * 2, HEIGHT * 2 });
     this->addChildAtPosition(m_bgSprite, Anchor::Center);
 
-    m_input = CCTextInputNode::create(width, HEIGHT, placeholder.c_str(), 24, font.c_str());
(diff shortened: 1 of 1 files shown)


loader/src/ui/nodes/TextInput.cpp
```diff
@@ -70,7 +70,7 @@ bool TextInput::init(float width, std::string const& placeholder, std::string co
     m_bgSprite->setContentSize({ width * 2, HEIGHT * 2 });
     this->addChildAtPosition(m_bgSprite, Anchor::Center);
 
-    m_input = CCTextInputNode::create(width, HEIGHT, placeholder.c_str(), 24, font.c_str());
+    m_input = CCTextInputNode::create(width-10.f, HEIGHT, placeholder.c_str(), 24, font.c_str());
```
matcool, line 73: can you add a space around the -  this is very important

FlafyDev: true

PR #983, opened 2024-06-30
New image arguments format

hiimjustin000:
Kept the old one for compatibility reasons

This has been tested


matcool:
it should be splitting at &, like if we were to support multiple args they would be `?scale=0.5&foo=12` not `?scale=0.5?foo=12`

hiimjustin000:
Aight then

hiimjustin000:
I fixed it, it's ready to merge now

PR #995, opened 2024-07-09
Update chinese translations

RayDeeUx:
per new strings added by fod


matcool:
(fod had no work in said strings)

PR #1019, opened 2024-07-28
Changed the operator overloads of CCPoint to be more inline with othe…

SMJSGaming:
…r engines


HJfod:
Context for anyone reviewing this in the future: this is technically an API break but the idea of a "less-than comparison" for 2d geometry doesn't make mathematical nor logical sense in the first place so I would be very surprised to find anyone relying on this behaviour beyond it making `CCPoint`s usable in maps

PR #1023, opened 2024-07-31
allow dashes in developer name search

hiimjustin000:
yes...


matcool:
maybe add _ too? what does github and or the server allow

qimiko:
is it worth handling the restriction that - can't appear at the start/end of usernames/next to another -

hiimjustin000:
i don't think so

PR #1029, opened 2024-08-04
Change SwelvyBG Sprites to be 2048x512

TheSillyDoggo:
Original sprite size was 2000x500 which is not a power of two, causing the sprite to not render on some devices (ios 💔)


matcool:
why would this fix anything? many many textures in gd are not even close to powers of 2

TheSillyDoggo:
GL_REPEAT causes textures to not render on some devices if the image size isn't a power of two

dankmeme01:
not sure which gles version GD is using, but the spec for gles 2.0 `glTexParameter` ([here](https://registry.khronos.org/OpenGL-Refpages/es2.0/xhtml/glTexParameter.xml), notes section) mentions that the width and height of the texture should be powers of two if the wrapping mode is not `GL_CLAMP_TO_EDGE`

PR #1040, opened 2024-08-13
Added IDs to web request to prevent incompatibilities between web mods

SMJSGaming:
Currently web mods need to check for pointers to prevent hook recursion. But to prevent the request from being thrown out you also need to copy it. This makes it so there can never be 2 web mods active at the same time.

### DIFF
--- a/loader/include/Geode/utils/web.hpp
+++ b/loader/include/Geode/utils/web.hpp
@@ -112,8 +112,9 @@ namespace geode::utils::web {
     private:
         class Impl;
 
+        static size_t m_idCounter;
         std::shared_ptr<Impl> m_impl;
-
+        const size_t m_id;
     public:
         WebRequest();
         ~WebRequest();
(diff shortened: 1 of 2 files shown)


loader/src/utils/web.cpp
```diff
@@ -212,7 +212,9 @@ class WebRequest::Impl {
     }
 };
 
-WebRequest::WebRequest() : m_impl(std::make_shared<Impl>()) {}
+size_t WebRequest::m_idCounter = 0;
+
+WebRequest::WebRequest() : m_impl(std::make_shared<Impl>()), m_id(WebRequest::m_idCounter++) {}
```
HJfod, line 217: This is a race condition, WebRequests aren't necessarily created in the same thread

loader/include/Geode/utils/web.hpp
```diff
@@ -112,8 +112,9 @@ namespace geode::utils::web {
     private:
         class Impl;
 
+        static size_t m_idCounter;
```
HJfod, line 115: This should be atomic_size_t

HJfod: and also static members should be prefixed with `s_` not `m_`

HJfod (changes requested):

HJfod (approved):

PR #1047, opened 2024-08-20
Fixing current working directory on startup

iliasHDZ:
The game crashes if launched from command prompt from a directory other than the game's directory.
This should fix the CWD back to the GD directory.

I need this so I can easily run GD from a batch script after compiling my mod.

### DIFF
--- a/loader/src/platform/windows/main.cpp
+++ b/loader/src/platform/windows/main.cpp
@@ -151,7 +151,6 @@ static void fixCWD() {
             break;
         }
     }
-    console::messageBox("Geode", fmt::format("Hello from gdMainHook!\Setting CWD: {}", cwd));
     SetCurrentDirectoryA(cwd);
 }
 
--- a/loader/src/platform/windows/main.cpp
+++ b/loader/src/platform/windows/main.cpp
@@ -138,10 +138,29 @@ void* mainTrampolineAddr;
 #include "gdTimestampMap.hpp"
 unsigned int gdTimestamp = 0;
 
+// In case the game is launched from a different directory through command line
+// this function will set the current working directory to the game's directory
+// to avoid the game crashing due to not being able to find the resources
+static void fixCWD() {
+    char cwd[1024];
+    DWORD size = GetModuleFileNameA(NULL, cwd, sizeof(cwd));
+    if (size == sizeof(cwd)) return;
+    for (int i = size - 1; i >= 0; i--) {
+        if (cwd[i] == '\\') {
+            cwd[i] = '\0';
+            break;
+        }
+    }
+    console::messageBox("Geode", fmt::format("Hello from gdMainHook!\Setting CWD: {}", cwd));
+    SetCurrentDirectoryA(cwd);
+}
+
 int WINAPI gdMainHook(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPWSTR lpCmdLine, int nCmdShow) {
     // MessageBoxA(NULL, "Hello from gdMainHook!", "Hi", 0);
 
     updateGeode();
+
+    fixCWD();
 
     if (versionToTimestamp(GEODE_STR(GEODE_GD_VERSION)) > gdTimestamp) {
         console::messageBox(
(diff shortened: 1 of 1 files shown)


matcool:
this will likely break on unicode paths due to the usage of the A winapi functions

loader/src/platform/windows/main.cpp
```diff
@@ -138,11 +138,29 @@ void* mainTrampolineAddr;
 #include "gdTimestampMap.hpp"
 unsigned int gdTimestamp = 0;
 
+// In case the game is launched from a different directory through command line
+// this function will set the current working directory to the game's directory
+// to avoid the game crashing due to not being able to find the resources
+static void fixCWD() {
+    char cwd[1024];
+    DWORD size = GetModuleFileNameA(NULL, cwd, sizeof(cwd));
```
matcool, line 146: use W methods instead

loader/src/platform/windows/main.cpp
```diff
@@ -138,11 +138,29 @@ void* mainTrampolineAddr;
 #include "gdTimestampMap.hpp"
 unsigned int gdTimestamp = 0;
 
+// In case the game is launched from a different directory through command line
+// this function will set the current working directory to the game's directory
+// to avoid the game crashing due to not being able to find the resources
+static void fixCWD() {
+    char cwd[1024];
```
matcool, line 145: MAX_PATH should be used instead of an arbitrary number

iliasHDZ: I was using a different number since i found max path to be rather small but sure

iliasHDZ:
So I fixed it. Will this be checked soon?

loader/src/platform/windows/main.cpp
```diff
@@ -138,11 +138,29 @@ void* mainTrampolineAddr;
 #include "gdTimestampMap.hpp"
 unsigned int gdTimestamp = 0;
 
+// In case the game is launched from a different directory through command line
+// this function will set the current working directory to the game's directory
+// to avoid the game crashing due to not being able to find the resources
+static void fixCWD() {
+    WCHAR cwd[MAX_PATH];
+    DWORD size = GetModuleFileNameW(NULL, cwd, sizeof(cwd));
```
altalk23, line 146: the sizeof here will not work since it will return maxpath * sizeof wchar, using a std array would make the much more easier

loader/src/platform/windows/main.cpp
```diff
@@ -138,11 +138,29 @@ void* mainTrampolineAddr;
 #include "gdTimestampMap.hpp"
 unsigned int gdTimestamp = 0;
 
+// In case the game is launched from a different directory through command line
+// this function will set the current working directory to the game's directory
+// to avoid the game crashing due to not being able to find the resources
+static void fixCWD() {
+    WCHAR cwd[MAX_PATH];
```
altalk23, line 145: we in geode refuse to use c related stuff as much as we can, and for good reason, using std::array here would be much better and would fix that bug below

loader/src/platform/windows/main.cpp
```diff
@@ -138,11 +138,29 @@ void* mainTrampolineAddr;
...
+    DWORD size = GetModuleFileNameW(NULL, cwd, sizeof(cwd));
+    if (size == sizeof(cwd)) return;
+    for (int i = size - 1; i >= 0; i--) {
```
altalk23, line 148: im not sure how i feel about this part since it will replace wrong if the string is smth like path\GeometryDash.exe which is bad

altalk23 (changes requested):

loader/src/platform/windows/main.cpp
iliasHDZ: i do understand it feels kinda wrong but this is the simplest solution i could come up with. I mean, i could convert it to `std::string` and then use `std::find`. But i don't want like half the file to be about fixing the CWD. I feel like this is short and concise.

iliasHDZ:
Ok, it now uses `std::array`. I also fixed some cases where for example there is no slashes, it will return.

PR #1060, opened 2024-09-01
Added a v2 of text area which is more reliable and has more options

SMJSGaming:


loader/include/Geode/ui/SimpleTextArea.hpp
```diff
@@ -22,7 +22,7 @@ namespace geode {
      * 
      * Contact me on Discord (\@smjs) if you have any questions, suggestions or bugs.
     */
-    class GEODE_DLL SimpleTextArea : public cocos2d::CCNode {
+    class GEODE_DLL [[deprecated("Use TextArea instead")]] SimpleTextArea : public cocos2d::CCNode {
```
matcool, line 25: do you mean textareav2?

SMJSGaming: Oh right, forgot to change that back

loader/include/Geode/ui/TextAreaV2.hpp
```diff
@@ -0,0 +1,241 @@
+#pragma once
+
+#include <cocos2d.h>
+#include "PaddingNode.hpp"
+
+namespace geode {
+    class GEODE_DLL TextAreaV2 : public PaddingNode {
```
HJfod, line 7: The naming convention for generic Geode UI classes is to give them short, reasonable names, ones that you would expect the given class to have in any UI framework (see `ScrollLayer`, `TextInput`, `ListView`, etc.) This should be called something analogous to `TextArea` like `MultilineLabel`, or possibly even just `Label` or `Text` (these would definitely need further discussion though)

loader/src/ui/nodes/TextAreaV2.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+        return area;
+    } else {
+        CC_SAFE_DELETE(area);
+
+        return nullptr;
+    }
+}
+
+geode::TextAreaV2::TextAreaV2(const std::string& text, const std::string& font, const float scale, const float width, const float height, const bool deferUpdates) :
```
HJfod, line 26: Cocos2d UI patterns do not ever use constructors to set members. Set them in `init`

loader/src/ui/nodes/TextAreaV2.cpp
```diff
@@ -0,0 +1,377 @@
...
+    PaddingNode(CCNode::create()),
```
HJfod, line 27: Default values that incur any explicit heap allocations should by convention always happen in `init` and not directly in the member

loader/src/ui/nodes/TextAreaV2.cpp
```diff
@@ -0,0 +1,377 @@
...
+    m_text(text),
+    m_font(font),
+    m_textColor({ 255, 255, 255, 255 }),
```
HJfod, line 30: Default values that do not incur a heap alloc should be set at the declaration using assignment initialization directly and not here

loader/src/ui/nodes/TextAreaV2.cpp
```diff
@@ -0,0 +1,377 @@
+#include <Geode/ui/TextAreaV2.hpp>
+#include <Geode/DefaultInclude.hpp>
+
+using namespace geode::prelude;
+
+#define IMPL_GETTER(type, name, methodName) \
```
HJfod, line 6: We generally don't use macros like this, but it's fine here ig

loader/src/ui/nodes/TextAreaV2.cpp
```diff
@@ -0,0 +1,377 @@
...
+    type geode::TextAreaV2::get##methodName() const { return m_##name; }
+#define IMPL_GETTER_SETTER(type, paramType, name, methodName) \
+    void geode::TextAreaV2::set##methodName(const paramType name) { m_##name = name; this->update(); } \
+    IMPL_GETTER(type, name, methodName)
+
+geode::TextAreaV2* geode::TextAreaV2::create(const std::string& text, const std::string& font, const float scale, const float width, const float height, const bool deferUpdates) {
+    TextAreaV2* area = new TextAreaV2(text, font, scale, width, height, deferUpdates);
```
HJfod, line 13: This should be `auto`

loader/src/ui/nodes/TextAreaV2.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+    if (area && area->init()) {
+        area->autorelease();
+
+        return area;
+    } else {
```
HJfod, line 19: This `else` is unnecessary

loader/src/ui/nodes/TextAreaV2.cpp
```diff
@@ -0,0 +1,377 @@
...
+    PaddingNode(CCNode::create()),
+    m_text(text),
+    m_font(font),
+    m_textColor({ 255, 255, 255, 255 }),
+    m_alignment(Alignment::Left),
+    m_wrappingMode(WrappingMode::WordWrap),
+    m_maxLines(-1),
+    m_textScale(scale),
+    m_linePadding(0),
+    m_minWidth(-1),
```
HJfod, line 36: If a value is optional and its null state isn't logically 0, it should be explicitly marked `std::optional`

HJfod: Aka the member should be `std::optional` and this should be `std::nullopt`

loader/src/ui/nodes/TextAreaV2.cpp
```diff
@@ -0,0 +1,377 @@
...
+    const bool upperBoundWidth = m_maxWidth >= 0;
+    const bool upperBoundHeight = m_maxHeight >= 0;
+    const bool artificialWidth = m_minWidth >= 0 || upperBoundWidth;
+    const bool artificialHeight = m_minHeight >= 0 || upperBoundHeight;
+    const bool inheritedNodeSize = this->getContentSize() == this->getPaddedContainerSize();
+    float height = m_lines.empty() ? 0 : m_lines.back().currentHeight;
+    float width = 0;
+
+    // First determine the container size before manipulating the node anchors and true positions
+    for (const Line& line : m_lines) {
```
HJfod, line 145: `auto`

loader/src/ui/nodes/TextAreaV2.cpp
```diff
@@ -0,0 +1,377 @@
...
+        width = std::max(width, line.label->getScaledContentWidth());
+    }
+
+    if (artificialWidth) {
+        width = std::max(m_minWidth, width);
+
+        m_container->setContentWidth(upperBoundWidth ? std::min(m_maxWidth, width) : width);
+    } else {
```
HJfod, line 153: Everyone else in the Geode team hates me for this so you can ignore this but `else` should be preceded by a newline

loader/src/ui/nodes/TextAreaV2.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+        // Correct the Y position to be relative to the container height
+        line.label->setPositionY(m_container->getContentHeight() - line.currentHeight + line.label->getScaledContentHeight());
+        m_container->addChild(line.label);
+    }
+}
+
+void geode::TextAreaV2::updateLineAlignment(CCLabelBMFont* line) {
+    switch (m_alignment) {
+        case Alignment::Left:
```
HJfod, line 180: case labels that aren't a single line should be braced i.e. `case N: {} break;`

loader/src/ui/nodes/TextAreaV2.cpp
```diff
@@ -0,0 +1,377 @@
...
+            line->setAnchorPoint({ 0, 1 });
+            line->setPositionX(0);
+            break;
```
HJfod, line 183: the `break` along with the brace mentioned above should be on the same indent as the `case` for legibility and there should be a newline afterwards

loader/src/ui/nodes/TextAreaV2.cpp
```diff
@@ -0,0 +1,377 @@
...
+        if ((character->getPositionX() - character->getContentWidth() / 2) * m_textScale <= m_maxWidth) {
+            return overflow;
+        }
+    }
+
+    return 0;
+}
+
+geode::TextAreaV2::Line geode::TextAreaV2::createLine(const std::string& text, Line& previousLine) {
+    CCLabelBMFont* line = CCLabelBMFont::create(text.c_str(), m_font.c_str());
```
HJfod, line 233: `auto`

loader/src/ui/nodes/TextAreaV2.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+    line->setScale(m_textScale);
+    line->setColor({ m_textColor.r, m_textColor.g, m_textColor.b });
+    line->setOpacity(m_textColor.a);
+
+    const size_t lineSize = text.size();
+    const float currentHeight = m_linePadding + previousLine.currentHeight + line->getScaledContentHeight();
```
HJfod, line 240: In cases like these not having `auto` is good because it isn't immediately obvious that the return value of a numeric computation is a `float`

loader/src/ui/nodes/TextAreaV2.cpp
```diff
@@ -0,0 +1,377 @@
...
+    const size_t overflow = this->isWidthOverflowing(line) ? this->getOverflowAmount(line, lineSize) : 0;
+    const size_t overflowStart = lineSize - overflow;
+    const std::string finalText = text.substr(0, overflowStart);
```
HJfod, line 243: Similarly in cases like strings `auto` is fair to avoid due to the multiple different string types

loader/src/ui/nodes/TextAreaV2.cpp
```diff
@@ -0,0 +1,377 @@
...
+    const bool exceededLines = (m_maxLines != -1 && previousLine.lineNumber + 1 >= m_maxLines) ||
```
HJfod, line 244: `bool` should also always be `bool` to avoid implicit conversion bugs

loader/src/ui/nodes/TextAreaV2.cpp
```diff
@@ -0,0 +1,377 @@
...
+            currentLine.label->setString(currentLine.text.c_str());
+        }
+
+        return this->createCutoffWrap(currentLine.overflow.erase(0, currentLine.overflow.find_first_not_of(' ')), currentLine);
+    });
+}
+
+std::vector<geode::TextAreaV2::Line> geode::TextAreaV2::createDelimitedWrap(const std::string& text, Line& reference, const std::string& delimiters) {
+    return this->wrapper(text, reference, [this, delimiters](Line& currentLine) {
+        if (currentLine.text.empty()) return std::vector<Line>();
```
HJfod, line 316: This should be braced and newline with indent

loader/include/Geode/ui/TextAreaV2.hpp
```diff
@@ -0,0 +1,241 @@
...
+        /**
+        * Sets whether the text should be hyphenated
+        */
+        void setHyphenate(const bool hyphenate);
+        /**
+        * Gets whether the text should be hyphenated
+        */
+        bool getHyphenate() const;
+        /**
+        * Sets whether the text area should show an ellipsis when the text overflows
```
HJfod, line 179: All these doc comments should have the stars aligned to the first star of the opening `/**` using a single space

loader/include/Geode/ui/TextAreaV2.hpp
```diff
@@ -0,0 +1,241 @@
...
+        */
+        void setText(const std::string& text);
+        /**
+        * Gets the text of the text area
+        */
+        std::string getText() const;
+        /**
+        * Sets the color of the text in the text area
+        */
+        void setTextColor(const cocos2d::ccColor4B& color);
```
HJfod, line 59: We use `const&`

HJfod: Also `ccColor4B` is pointer-size and should be passed by value rather than `const&` anyway

loader/include/Geode/ui/TextAreaV2.hpp
```diff
@@ -0,0 +1,241 @@
...
+        /**
+        * Gets the color of the text in the text area
+        */
+        cocos2d::ccColor4B getTextColor() const;
+        /**
+        * Sets horizontal the alignment of the text in the text area
+        */
+        void setAlignment(const Alignment alignment);
```
HJfod, line 67: This is fine because it's not a ref, though the `const` here is also utterly redundant

loader/include/Geode/ui/TextAreaV2.hpp
```diff
@@ -0,0 +1,241 @@
...
+        enum class Alignment {
+            // Aligns the text to the left
+            Left,
+            // Aligns the text in the center
+            Center,
+            // Aligns the text to the right
+            Right
+        };
+
+        struct Line {
```
HJfod, line 29: This should be PIMPL'd, or in some other way not be exposed in a header! Guaranteed ABI break in the future!

loader/include/Geode/ui/PaddingNode.hpp
```diff
@@ -0,0 +1,133 @@
...
+#include <cocos2d.h>
+
+namespace geode {
+    class GEODE_DLL PaddingNode : public cocos2d::CCNode {
+    public:
+        enum class PaddingType {
+            // The container will have its size adjusted to fit the padding
+            Contained,
+            // This node will have its size adjusted to fit the padding
+            Uncontained
```
HJfod, line 12: Should have trailing comma

loader/include/Geode/ui/PaddingNode.hpp
```diff
@@ -0,0 +1,133 @@
...
+            operator float() const;
+            operator cocos2d::CCPoint() const;
+
+            float top;
+            float right;
+            float bottom;
+            float left;
+
+            Padding();
+            Padding(cocos2d::CCPoint padding);
```
HJfod, line 25: `CCPoint` is not pointer-size on all platforms (Android32) so this should be `const&`

HJfod (changes requested):
coughing smjs pr baby vs hydrogen hjfod review bomb

loader/src/ui/nodes/TextAreaV2.cpp
SMJSGaming: Thank god, writing them all out would be painful

loader/src/ui/nodes/TextAreaV2.cpp
SMJSGaming: I opted to not do this since there's frequent math usage involved with these making it hell to use if they're optional

loader/src/ui/nodes/TextAreaV2.cpp
SMJSGaming: Hard agree, but that's usually a hot take so I didn't do it here for that reason

loader/include/Geode/ui/TextAreaV2.hpp
SMJSGaming: huh, I did that but I guess vscode auto formatted it

loader/src/ui/nodes/TextAreaV2.cpp
```diff
@@ -0,0 +1,377 @@
...
+#define IMPL_GETTER(type, name, methodName) \
+    type geode::TextAreaV2::get##methodName() const { return m_##name; }
+#define IMPL_GETTER_SETTER(type, paramType, name, methodName) \
+    void geode::TextAreaV2::set##methodName(const paramType name) { m_##name = name; this->update(); } \
+    IMPL_GETTER(type, name, methodName)
+
+geode::TextAreaV2* geode::TextAreaV2::create(const std::string& text, const std::string& font, const float scale, const float width, const float height, const bool deferUpdates) {
+    TextAreaV2* area = new TextAreaV2(text, font, scale, width, height, deferUpdates);
+
+    if (area && area->init()) {
```
matcool, line 15: if (area) is pointless, area will never be nullptr
either use std::nothrow or dont bother to check it (geode does the latter)

SMJSGaming:
I do not want this to be used by geode. Ever.

PR #1066, opened 2024-09-07
Split update checks into multiple batches

qimiko:
people have reached the maximum amount of mods that the servers can handle (idk why...), so the update check silently fails.
this resolves the first issue, which is that the update check fails. it doesn't fix the silent part

also this is really hacky bc i wanted to chain the tasks instead of running them all at once so i'm not actually sure if this works on anything not macos


Cvolton:
works on windows officially confirmed (official)

matcool (approved):

PR #1075, opened 2024-09-10
add easteregg for the jesus mod

TheBearodactyl:
now, if there's an error caused by the jesus mod, instead of using the word `its` to refer to the mod, it uses the word `His` (with the capital H)

ik this will probably be denied but it's worth a shot


iliasHDZ:
lmao

PR #1080, opened 2024-09-13
Fix Mac ObjC/++ autobuilds

MarioMastr:
When using the autobuild scripts for building MacOS geode projects, and said geode project includes an Objective C or C++ file, the GitHub runner throws an error regarding the languages being disabled in the pch file but being now currently enabled. This is the fix as implemented according to https://github.com/cocos2d/cocos2d-x/issues/20607#issuecomment-780266298, which _should_ solve this, however I don't think I have a way to actually test it. Worth a shot, though.


altalk23:
no, you are supposed to manually disable precompiled headers for objc files. geode already does this. see https://github.com/geode-sdk/geode/blob/4fa7d926e81a5afe5b45192a30a5f1ff73f62d47/loader/CMakeLists.txt#L94

PR #1099, opened 2024-10-01
revert "remove this thing"

FireMario211:
this broke wine when you try launching gd without using a terminal
image

this reverts commit https://github.com/geode-sdk/geode/commit/264cb28d882fe8b4aa4a12da7e91b061f1799f07


dankmeme01:
don't merge pls, this will break windows

FireMario211:
> don't merge pls, this will break windows

hmm i think i see the issue!

PR #1108, opened 2024-10-08
Settings + some list fixes

SMJSGaming:


loader/include/Geode/ui/General.hpp
```diff
@@ -86,7 +86,7 @@ namespace geode {
     public:
         static ListBorders* create();
 
-        void setSpriteFrames(const char* topAndBottom, const char* sides, float topPadding = 7.5f);
+        void setSpriteFrames(const char* topAndBottom, const char* sides, float horizontalPadding = 7.5f);
```
matcool, line 89: why the rename?

SMJSGaming: look into how it's used
it's used as both top and bottom padding and ends up being used as horizontal padding for the top and bottom sprite

loader/src/ui/nodes/TextInput.cpp
```diff
@@ -151,7 +151,6 @@ void TextInput::setCallback(std::function<void(std::string const&)> onInput) {
     m_onInput = onInput;
 }
 void TextInput::setEnabled(bool enabled) {
-    m_input->setMouseEnabled(enabled);
```
SMJSGaming, line 154: To clarify, this caused the scroll issues in all lists and text input doesn't even use the mouse besides touches

loader/src/ui/mods/GeodeStyle.cpp
```diff
@@ -217,19 +222,18 @@ std::string geodeTagName(std::string_view tag) {
...
-ListBorders* createGeodeListBorders(CCSize const& size) {
+ListBorders* createGeodeListBorders(CCSize const& size, bool forceDisableTheme) {
     auto ret = ListBorders::create();
-    if (isGeodeTheme()) {
+    const bool geodeTheme = isGeodeTheme(forceDisableTheme);
+    if (geodeTheme) {
         ret->setSpriteFrames("geode-list-top.png"_spr, "geode-list-side.png"_spr, 2);
+        ret->setContentSize(size);
+    } else {
+        ret->setContentSize(size + ccp(5, 5));
```
SMJSGaming, line 232: To explain this real quick, blame the fact that 90% of this asset is padding

GJ_commentTop_001

loader/src/ui/mods/GeodeStyle.cpp
```diff
@@ -74,12 +74,17 @@
     });
 }
 
-bool GeodeSquareSprite::init(CCSprite* top, bool* state) {
-    if (!CCSprite::initWithFile(isGeodeTheme() ? "GE_button_05.png"_spr : "GJ_button_01.png"))
+bool isGeodeTheme(bool forceDisableTheme) {
```
HJfod, line 77: This should be marked `static`

loader/src/ui/mods/GeodeStyle.cpp
HJfod: We love hardcoded values for dealing with weird textures

loader/src/ui/nodes/TextInput.cpp
HJfod: If `setTouchEnabled` is enough to make the input non-clickable then this is fine by me

HJfod (approved):

loader/src/ui/mods/GeodeStyle.cpp
matcool: should it though? its in the header

PR #1148, opened 2024-11-14
Fix mod downloads not using the metadata version to download

Fleeym:
Currently, mods fetch mod metadata from /v1/mods/{id}/latest version, if no version is passed in ModDownloadManager::startDownload(). 

At the same time, this request doesn't include the current GD version as a query param, so it can (and will) fetch mods from incompatible game versions.

Fix is adding this query param as a failsafe and using the mod metadata to download, wherever possible.

We also have to backport this fix to v3 quite fast

### DIFF
--- a/loader/src/server/Server.cpp
+++ b/loader/src/server/Server.cpp
@@ -8,6 +8,7 @@
 #include <fmt/chrono.h>
 #include <loader/LoaderImpl.hpp>
 #include "../internal/about.hpp"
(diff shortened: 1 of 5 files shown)


Cvolton:
lgtm (i didnt test it)

Cvolton (approved):

Fleeym:
No idea how to test tbh

Fleeym:
image
Tested fix on 2.206, works well

PR #1150, opened 2024-11-14
Updated the headers map in WebRequest and WebResponse to accept multiple headers with the same name

B1rtek:
Because there can be multiple headers with the same name, for example Set-Cookie (which was the main motivation for this change)
Also it's probably terrible idk (although the game didn't explode when I tried to use it and it actually worked)

### DIFF
--- a/loader/include/Geode/utils/web.hpp
+++ b/loader/include/Geode/utils/web.hpp
@@ -83,6 +83,14 @@ namespace geode::utils::web {
 
         std::vector<std::string> headers() const;
         std::optional<std::string> header(std::string_view name) const;
+
+        /**
+         * Retrieves a list of all headers from the response with a given name - there can be
+         * multiple headers with the same name, such as Set-Cookie, with each cookie in a separate
+         * header
+         * @param name name of the header
+         * @return std::optional<std::vector<std::string>>
+         */
         std::optional<std::vector<std::string>> headersWithName(std::string_view name) const;
     };
 
@@ -287,7 +295,7 @@ namespace geode::utils::web {
         /**
          * Gets the request headers
          *
-         * @return std::unordered_map<std::string, std::string>
+         * @return std::unordered_map<std::string, std::vector<std::string>>
          */
         std::unordered_map<std::string, std::vector<std::string>> getHeaders() const;
 
--- a/loader/include/Geode/utils/web.hpp
+++ b/loader/include/Geode/utils/web.hpp
@@ -91,7 +91,7 @@ namespace geode::utils::web {
          * @param name name of the header
          * @return std::optional<std::vector<std::string>>
          */
-        std::optional<std::vector<std::string>> headersWithName(std::string_view name) const;
+        std::optional<std::vector<std::string>> getAllHeadersNamed(std::string_view name) const;
     };
 
     class GEODE_DLL WebProgress final {
--- a/loader/include/Geode/utils/web.hpp
+++ b/loader/include/Geode/utils/web.hpp
@@ -82,7 +82,7 @@ namespace geode::utils::web {
         Result<> into(std::filesystem::path const& path) const;
 
         std::vector<std::string> headers() const;
-        std::optional<std::string> header(std::string_view name) const;
+        std::optional<std::string> header(std::string_view name, int index = 0) const;
     };
 
     class GEODE_DLL WebProgress final {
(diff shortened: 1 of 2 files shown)


loader/include/Geode/utils/web.hpp
```diff
@@ -82,7 +82,7 @@ namespace geode::utils::web {
         Result<> into(std::filesystem::path const& path) const;
 
         std::vector<std::string> headers() const;
-        std::optional<std::string> header(std::string_view name) const;
+        std::optional<std::string> header(std::string_view name, int index = 0) const;
```
Fleeym, line 85: If we're doing this change imo this should return the full vector

Fleeym (changes requested):

loader/include/Geode/utils/web.hpp
B1rtek: I did it in this way to make it backwards compatible but yeah that sounds like a better solution

matcool: personally id rather have a new method, since for 99% of use cases the header value you want is just a single value

B1rtek: ok, header(sv) returns a single header like it used to after a change I made, and I added headersWithName(sv) that returns the whole vector instead

loader/src/utils/web.cpp
```diff
@@ -406,7 +414,14 @@ WebTask WebRequest::send(std::string_view method, std::string_view url) {
...
                     value = value.substr(0, value.size() - 1);
                 }
-                headers.insert_or_assign(key, value);
+                // Create a new vector and add to it or add to an already existing one
+                if (headers.contains(key)) {
+                    headers.at(key).push_back(value);
+                } else {
+                    std::vector<std::string> headerValues;
+                    headerValues.push_back(value);
+                    headers.insert_or_assign(key, headerValues);
```
matcool, line 423: should probably move the vector instead of copying here

loader/src/utils/web.cpp
```diff
@@ -508,7 +523,16 @@ WebRequest& WebRequest::header(std::string_view name, std::string_view value) {
...
-    m_impl->m_headers.insert_or_assign(std::string(name), std::string(value));
+    // Create a new vector and add to it or add to an already existing one
+    std::string strName = std::string(name);
+    std::string strValue = std::string(value);
+    if (m_impl->m_headers.contains(strName)) {
+        m_impl->m_headers.at(strName).push_back(strValue);
+    } else {
+        std::vector<std::string> headerValues;
+        headerValues.push_back(strValue);
+        m_impl->m_headers.insert_or_assign(strName, headerValues);
```
matcool, line 534: same here

loader/src/utils/web.cpp
Fleeym: you can also just construct it in place

loader/include/Geode/utils/web.hpp
```diff
@@ -288,7 +289,7 @@ namespace geode::utils::web {
          *
          * @return std::unordered_map<std::string, std::string>
```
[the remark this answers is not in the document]
B1rtek: updated in the new commit

loader/include/Geode/utils/web.hpp
```diff
@@ -83,6 +83,7 @@ namespace geode::utils::web {
 
         std::vector<std::string> headers() const;
         std::optional<std::string> header(std::string_view name) const;
+        std::optional<std::vector<std::string>> headersWithName(std::string_view name) const;
```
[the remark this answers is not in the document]
B1rtek:
```cpp
/**
         * Retrieves a list of all headers from the response with a given name - there can be
         * multiple headers with the same name, such as Set-Cookie, with each cookie in a separate
         * header
         * @param name name of the header
         * @return std::optional<std::vector<std::string>>
         */
```

loader/src/utils/web.cpp
B1rtek: done

Fleeym:
So as a little personal recap:

- m_headers is now `std::unordered_map<std::string, std::vector<std::string>>`
- you can still use res->header(name), and if the key has multiple headers it returns the first one(?)
- you have res->headersWithName(name), which returns the full vector of values under that header key (vec can be empty)

if that's all then this lgtm, but maybe we could change `headersWithName` to something else, current name just doesn't sound right

Fleeym (approved):

B1rtek:
I renamed it to `getAllHeadersNamed()`

B1rtek:
:tada:

PR #1157, opened 2024-11-16
Add OverlayManager

TheSillyDoggo:
Adds a general notification node manager for multiple mods to easily add overlays that get drawn last, over scene transitions and the fps counter.


altalk23:
persisting across scenes does the same

TheSillyDoggo:
notification node draws over scene transitions, persist between scenes does not.

PR #1158, opened 2024-11-16
Added pt-pt

OmgRod:
portuguese portuguese


Cvolton:
finally the better portuguese

PR #1175, opened 2024-11-22
implement `Task.hpp` TODOs

TheBearodactyl:
- ~~added a progress mapper (line 860)~~
- implement `unhandled_exception` (line 969)
- put tasks into their own pools (line 885)
- ~~guarantee main thread (line 942)~~

i made sure to not make any changes to `.clang-format`, so any formatting changes are unintentional

### DIFF
--- a/loader/src/ui/mods/list/ModItem.cpp
+++ b/loader/src/ui/mods/list/ModItem.cpp
@@ -225,6 +225,7 @@ bool ModItem::init(ModSource&& source) {
         },
         [this](server::ServerModMetadata const& metadata) {
             m_badgeContainer = CCNode::create();
+            m_badgeContainer->setID("badge-container");
             m_badgeContainer->setLayoutOptions(AxisLayoutOptions::create()->setScaleLimits(.1f, .8f));
             if (metadata.featured) {
                 m_badgeContainer->addChild(CCSprite::createWithSpriteFrameName("tag-featured.png"_spr));


loader/include/Geode/utils/Task.hpp
```diff
@@ -882,23 +948,25 @@ namespace geode {
             return ListenerResult::Propagate;
         }
 
-        // todo: i believe alk wanted tasks to be in their own pool
         EventListenerPool* getPool() const {
-            return DefaultEventListenerPool::get();
+            static EventListenerPool* taskPool;
```
matcool, line 952: thats not-

TheBearodactyl: bruh

loader/include/Geode/utils/Task.hpp
```diff
@@ -1016,7 +1097,10 @@ namespace geode {
                     static_cast<void*>(new EventListener<Task<T, P>>(
                         [handle](auto* event) {
                             if (event->getValue()) {
-                                handle.resume();
+                                // resume on main thread
+                                queueInMainThread([handle]() {
```
matcool, line 1101: this is not needed, event listener callback is already guaranteed to be on the main thread

TheBearodactyl: fair enough

TheBearodactyl: fixed

loader/include/Geode/utils/Task.hpp
TheBearodactyl: idk why it did that
fixed

loader/include/Geode/utils/Task.hpp
```diff
@@ -963,10 +1030,23 @@ namespace geode {
...
-            std::suspend_never initial_suspend() noexcept { return {}; }
-            std::suspend_never final_suspend() noexcept { return {}; }
-            // TODO: do something here?
-            void unhandled_exception() {}
+            std::suspend_never initial_suspend() noexcept {
+                // Queue the coroutine to start on main thread
+                queueInMainThread([this]() {
+                    std::coroutine_handle<TaskPromise>::from_promise(*this).resume();
+                });
+                return {};
```
matcool, line 1038: does this even work? youre returning suspend_never which means the coroutine will not be blocked

loader/include/Geode/utils/Task.hpp
```diff
@@ -825,39 +888,42 @@ namespace geode {
...
+                                        static_cast<EventListener<NewTask>*>(ptr)->getFilter().cancel();
+                                    }
+                                );
                         }
                         else if (auto p = event->getProgress()) {
-                            // no guarantee P and NewProgress are compatible
-                            // nor does it seem like the intended behavior?
-                            // TODO: maybe add a mapper for progress?
+                            // Map progress from P to NewProgress if possible
+                            if constexpr (std::is_convertible_v<P, NewProgress>) {
```
Fleeym, line 924: This can lead to weird behaviour.

Let's say we're chaining 2 web requests.

If I understand this correctly, then progress will roll up to 100, then back to 0 and up to 100 again. Not good.

TheBearodactyl: how could i fix that?

Fleeym: plan here was to also create a mapper for progress, though there's no easy solution here. most chaining implementations don't care about the progress of the tasks before the final one.

loader/include/Geode/utils/Task.hpp
TheBearodactyl: my mistake

TheBearodactyl: @matcool would a suspend_always work?

loader/include/Geode/utils/Task.hpp
TheBearodactyl: lemme try something

TheBearodactyl: @Fleeym would something like this work?

```cpp
if constexpr (std::is_convertible_v<P, NewProgress>) {
    auto scaledProgress = static_cast<float>(*p) * 0.5f;
    NewTask::progress(handle.lock(), NewProgress(scaledProgress));
}
```

matcool: the progress doesnt have to be a number, but also you can chain more than 2 tasks

TheBearodactyl: @Fleeym what about something like this? probably not since i literally just came up with it but i might be able to get something from it

```cpp
             if constexpr (std::is_convertible_v<P, NewProgress>) {
                                size_t totalTasks = 1;
                                auto currentHandle = handle.lock();
                                while (currentHandle && currentHandle->m_extraData) {
                                    totalTasks++;
                                    auto listener = static_cast<EventListener<NewTask>*>(currentHandle->m_extraData->m_data);
                                    currentHandle = listener->getFilter().m_handle;
                                }
                                
                                auto scaledProgress = static_cast<float>(*p) / static_cast<float>(totalTasks);
                                NewTask::progress(handle.lock(), NewProgress(scaledProgress));
                            }
                    }
```

TheBearodactyl: find a way to count how many tasks are happening at once, then scale the progress tracker to that amount so that it doesn't restart the tracker every time a task ends

Fleeym: well progress isn't guaranteed to be a float, so this can't really work.

progress can be anything from an enum to a primitive type to a full blown class

TheBearodactyl: oh
ok yeah i see what you mean by there's no easy solution

TheBearodactyl: gimme about 2 hours and i'll see if i can come up with something, otherwise i'll just revert it to its original state

loader/include/Geode/utils/Task.hpp
```diff
@@ -882,7 +948,6 @@ namespace geode {
             return ListenerResult::Propagate;
         }
 
-        // todo: i believe alk wanted tasks to be in their own pool
```
matcool, line 885: dont remove the comment

TheBearodactyl: but wouldn't a todo for something that's implemented be kinda useless...?

matcool: its not implemented.

loader/include/Geode/utils/Task.hpp
matcool: assuming you want to linearly interpolate between enum values is crazy 😭 this feature seems overkill

TheBearodactyl: > assuming you want to linearly interpolate between enum values is crazy 😭 this feature seems overkill

i mean this pr is meant to implement all todos and this was something that was in a todo so

matcool: maybe some todos are not worth it, like explained in the comment itself and flame's msg

loader/include/Geode/utils/Task.hpp
TheBearodactyl: it will be once my [question](https://github.com/geode-sdk/geode/pull/1175#discussion_r1854919162) is answered

loader/include/Geode/utils/Task.hpp
TheBearodactyl: but i don't see the problem in implementing this, especially considering the fact that my proposed solution works with (as Fleeym said) primitives, enums, and classes?

matcool: is it really worth it when the behavior can be so arbitrary? (and also doing the task chain count is hacky + expensive)

loader/include/Geode/utils/Task.hpp
TheBearodactyl: i have an implementation i just need you to answer the question i asked like 4 hours ago

loader/include/Geode/utils/Task.hpp
TheBearodactyl: > is it really worth it when the behavior can be so arbitrary? (and also doing the task chain count is hacky + expensive)

one of the problems fleeym had with my original implementation was that the progress would reset for each task, the task count is just so that i can scale the progress tracker

Fleeym: In my opinion we should leave this as is, as changing this means an API break. Also the more obvious functionality here is just tracking the progress of the last task.

We'll keep in mind the implementation for a future major version though.

loader/include/Geode/utils/Task.hpp
Fleeym: At the moment your PR has the same implementation as we have currently, so please keep the comment in.

loader/include/Geode/utils/Task.hpp
TheBearodactyl: oki

loader/include/Geode/utils/Task.hpp
TheBearodactyl: comment is back

PR #1205, opened 2025-01-01
Make strings less confusing

PizzaTowerFanGD:
Yep.


PizzaTowerFanGD:
oh i meant the "Show Errors Only" and "Hide Errors Only" (as well as updates)

STGamer24YT:
I like these changes
This should get approved

PizzaTowerFanGD:
unfortunately this doesnt work because i used patch-1 instead of main.

Kingminer7:
not all errors mean broken

> oh i meant the "Show Errors Only" and "Hide Errors Only" (as well as updates)

not all errors mean something is broken

Fleeym:
Not sure this is the right way to change the strings, though I'll admit the wording here is strange.

Maybe we could keep the same string: "Updates Only" / "Errors Only" and only change the color of the button (might be bad for accessibility though)

PR #1211, opened 2025-01-08
Implement scrollable descriptions

RayDeeUx:
This should hopefully put an end to my history with hooking FLAlertLayer::init at -2123456789 priority.

Parses the `scroll-desc` JSON key to determine if the description popup should be scrollable.

### DIFF
--- a/loader/src/loader/SettingNodeV3.cpp
+++ b/loader/src/loader/SettingNodeV3.cpp
@@ -142,12 +142,19 @@ void SettingNodeV3::updateState(CCNode* invoker) {
 
 void SettingNodeV3::onDescription(CCObject*) {
     auto title = m_impl->setting->getDisplayName();
-    FLAlertLayer::create(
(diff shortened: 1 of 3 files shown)


loader/src/loader/SettingNodeV3.cpp
```diff
@@ -142,13 +142,20 @@ void SettingNodeV3::updateState(CCNode* invoker) {
...
     auto title = m_impl->setting->getDisplayName();
-    FLAlertLayer::create(
+    if (!m_impl->setting->getScrollableDescription()) FLAlertLayer::create(
         nullptr,
         title.c_str(),
         m_impl->setting->getDescription().value_or("No description provided"),
         "OK", nullptr,
         clamp(title.size() * 16, 300, 400)
     )->show();
+    else FLAlertLayer::create(
```
matcool, line 152: add some braces around if

Fleeym:
zmx created her own implementation with #1286, we'll use that one, but thanks for the idea!

PR #1213, opened 2025-01-12
Add 2.2 macOS UUIDs

hiimjustin000:


qimiko:
thank you for doing my work for me :slightly_smiling_face: (the pre 2.207 uuids aren't necessary in any way but i won't complain about having them)

PR #1214, opened 2025-01-12
Mark all cocos members public

Prevter:
Made sure all methods remain at their original protection level.
Also changed a few std::sets to gd::sets.


SpaghettDev:
this pr is so.... inflated

PR #1216, opened 2025-01-13
Update PlatformDetect.cmake

Gamer069:
Edited platform detection cmake script to make it work on linux


Gamer069:
very small fix, but good for qol

matcool:
why would it be set to Win64 on linux?

Gamer069:
> why would it be set to Win64 on linux?

because it's done via wine, isn't it? when you compile a mod, it adds it to gd. Gd itself can only be ran with proton/wine

Gamer069:
> to compile on linux you must set up a cross compilation toolchain, which will already properly set the WIN32 cmake variable. forcing it like this won't make it magically work

sorry, i didn't notice that. I'll close this pr now.

PR #1241, opened 2025-02-10
Add CCGLProgram::compileShader patches to Android and macOS

hiimjasmine00:
Mat told me to use addresser, but CCGLProgram::compileShader is a private function


qimiko:
despite not being able to access private functions in c++, [dlsym](https://man7.org/linux/man-pages/man3/dlsym.3.html) would be happy to give you the address to it

qimiko:
oh nvm i saw you did that 😭

PR #1246, opened 2025-02-20
Add `setSettingName` and `setSettingDescription` methods to `Mod`

TheBearodactyl:
v2 bc neovim fucked with formatting last time (used notepad++ this time)


jtmcb16 (approved):

HJfod:
Why does this add methods to `Mod` directly? What's the use case for this? Does it warrant having a quick-access method in `Mod`?

iAndyHD3 (approved):

TheBearodactyl:
updated my fork

TheBearodactyl:
> Why does this add methods to `Mod` directly? What's the use case for this? Does it warrant having a quick-access method in `Mod`?

well for one they're companion methods to the `get` methods so it'd be kinda weird NOT to have them directly in `Mod`. like imagine if a class had a getter for `foo` and the setter was in a completely different class for no reason. it'd just be unnecessarily complicated opposed to having both the getter and setter in the same class.
and the reason they exist is to make customizing setting menus easier (e.g. if someone makes a localization mod and they want to add support for setting menus and not just the base game)

iAndyHD3 (approved):

HJfod:
If this is just for localization, I'd rather add a dedicated localization API that'd let people provide a JSON file or something with translations listed. Allowing changing setting names and descriptions arbitarily at runtime seems like a bad move that'll confuse users when suddenly their setting has a different name.

If we do add this, I will say there should not be setters in `Mod`. Yes the getters are there, but the getters are meant to be used much much more frequently than a setter would be, and for the usecases provided I think people can write two more lines of code to set the value. Having it be in `Mod` directly both bloats its interface and makes it more encouraging to misuse.

TheBearodactyl:
fair enough. still gonna keep this open though until said localization api is a thing though

PR #1247, opened 2025-02-20
Update the pugixml headers to be v1.15 compatible

SMJSGaming:
This will prevent undefined errors when iterating over nodes or attributes in your mod with pugixml v1.15


altalk23:
piss baby

PR #1251, opened 2025-02-25
Add some GD textures

OmgRod:
Use the ones you want :D


matcool:
personally a little out of scope for geode

PR #1253, opened 2025-02-27
MacOS Microphone Permissions

LimeGradient:
On macOS and on game launch, this will check the Info.plist for the NSMicrophoneUsageDescription tag, if it is not found it will add it to the plist. This also implements the requestPermission method for Microphone on macOS.


altalk23:
would be better if checkPermissionPlist() existed in all platforms (and maybe renamed to checkPermissions) and they are empty in other platforms instead of having an ifdef otherwise lgtm

PR #1254, opened 2025-02-28
Made language in installer slightly more accessible (changed EnglishExtra.nsh, SpanishExtra.nsh, and SpanishInternationalExtra.nsh)

STGamer24YT:
- Changed the language to make it more accessible to users that aren't used to refer to folders with the word "path" (this is for consistency with some of the text the installer already displays)
image

- Fixed a small error in SpanishExtra.nsh: In spanish, we use the `¡` symbol to indicate the start of exclamation (and `¿` for the start of a question) and this file was missing 2 of them

I know these aren't really big changes, but I wanted to make them because the previous versions seemed weird to me.


B4uti4github:
You're right, I speak Spanish, and I understand it.

PR #1255, opened 2025-03-01
Add gnustl vector and map (and a lot of other dependencies those have)

0x5abe:
It's a lot of new code that I copied from the gcc libstdc++v3 at the first commit of 2014. I don't expect this to get merged soon but if someone else cares about this and bothers to check nothing breaks because of it then I'm all for it getting merged. Some mods like mine (PlatformerSaves/PersistenceAPI) rely on the memory structure of these stl types and until robtop decides to update the ndk the game uses this is the next best thing I guess. Making it a draft PR cause there's a bunch of little hacks and just in general stuff I scrambled together to get it to compile at all, and I just tested for a bit that geode launched properly with this and played for a bit.


matcool:
how does this differ from the existing gnustl map implementation?

0x5abe:
> how does this differ from the existing gnustl map implementation?

Building PersistenceAPI (my mod) with the old implementation for vector/map resulted in random crashes that appeared to be memory corruption. I wasn't able to pinpoint the exact cause but using this "official" gnustl implementation fixed those.

dankmeme01 (approved):
Tested, did not compile initially (when using map::emplace) but after some fixing everything seems to work fine now.

@0x5abe Would be preferred if you could test whether your mod compiles and works fine after my commit

0x5abe:
I'm getting these errors now:
image

I'm guessing cause now all of the containers are using the custom pair struct and if you try to access it and assign it to an std::pair it blows up. I think the best solution may be to add a conversion operator to std::pair from geode::stl::pair, but I'm having trouble with doing that rn. Another solution would be to have gd::pair like for the containers but this seems like it could break some mods apart from mine?

0x5abe:
This is the code that it errors on btw
image

SpaghettDev:
or you could just, ... `for (const auto& pair : pairs)` 👍

0x5abe:
Or things could be done the right way

matcool:
i assume the mysterious `i_value` is a `gd::map`? if so, its not too bad to assume that it wont be convertible to `std::pair` on iteration IMO

0x5abe:
Just pushed what I believe to be a good solution (convert properly from the custom pair type defined to std::pair).

loader/include/Geode/c++stl/gnustl/stl_pair.h
```diff
@@ -126,6 +126,11 @@ _GLIBCXX_BEGIN_NAMESPACE_VERSION
       constexpr pair(const pair&) = default;
       constexpr pair(pair&&) = default;
 
+    template<class _U1, class _U2>
+    operator std::pair<_U1, _U2>() {
```
SpaghettDev, line 130: why redefine the template types? why not reuse _T1 and _T2?

0x5abe: Uh cause I copy pasted from a diff place and wasn't really paying much attention, ur right it should just use the existing ones

0x5abe: Should be better now

0x5abe:
> > Or things could be done the right way
> 
> You are iterating over a struct from gnustl, using our `std::pair` is not the "right" way. I would've left the std::pair because it's functionally identical but it was causing issues with the allocator when using map::emplace so I replaced all tuples and pairs with `geode::stl::xxx`

I know using std::pair isn't the right way **now**, still some mods other than mine might have used gd:: containers assuming the pair types to be std::pair as well, since it was that way before for android, and it still is for other non android platforms. By doing it the right way I mean not breaking compatibility and forcing people to change their code when geode::stl::pair can perfectly be converted to std::pair like I ended up doing in the fix.

PR #709, opened 2024-04-20
Add handling of unused keys and mouse buttons, rewrite numpad input

Fleeym:
This PR implements unused keys and mouse buttons. It does not include mouse left click, right click, scroll click.
Additionally, it remaps numpad input from their KEY_1 - KEY_0 counterparts to their actual numpad enumKeyCodes.

Added key codes are as follows

```cpp
KEY_GraveAccent = 0x1000,
KEY_OEMEqual = 0x1001,
KEY_LeftBracket = 0x1002,
KEY_RightBracket = 0x1003,
KEY_Backslash = 0x1004,
KEY_Semicolon = 0x1005,
KEY_Apostrophe = 0x1006,
KEY_Slash = 0x1007,
KEY_Equal = 0x1008,
KEY_NumEnter = 0x1009,
// Keys used by some non-US keyboard layouts
KEY_World1 = 0x100A,
KEY_World2 = 0x100B,
// Mouse buttons (excluding clicks)
MOUSE_4 = 0x1100,
MOUSE_5 = 0x1101,
MOUSE_6 = 0x1102,
MOUSE_7 = 0x1103,
MOUSE_8 = 0x1104
```

`CCKeyboardDispatcher::dispatchKeyboardMSG` will use these codes instead of -1. Additionally, I have hooked `CCKeyboardDispatcher::keyToString` for the newly created key codes.

This will remain a draft PR until I figure out how MacOS does keyboard and mouse input.

PR #1258, opened 2025-03-04
Add handling of unused keys and mouse buttons, rewrite numpad input v2

SpaghettDev:
Spiritual successor to #709. Adds Mac support. Extends Android keys.


loader/include/Geode/cocos/robtop/keyboard_dispatcher/CCKeyboardDelegate.h
```diff
@@ -6,199 +6,220 @@
...
+        KEY_World1 = 0x100A,
+        KEY_World2 = 0x100B,
+
+        // Mouse buttons (excluding clicks)
+        MOUSE_4 = 0x1100,
+        MOUSE_5 = 0x1101,
+        MOUSE_6 = 0x1102,
+        MOUSE_7 = 0x1103,
+        MOUSE_8 = 0x1104
+    } enumKeyCodes;
```
matcool, line 222: why did you indent this whole enum

SpaghettDev: to match file indentation

SpaghettDev: alright

loader/src/hooks/AddExtraKeys.cpp
```diff
@@ -0,0 +1,206 @@
+#ifdef GEODE_IS_WINDOWS
```
matcool, line 1: relying on pch 💔 silly flame

qimiko:
you forgot the android inputs in `[loader/src/platform/android/input.cpp](https://github.com/geode-sdk/geode/blob/main/loader/src/platform/android/input.cpp)` :(

loader/src/hooks/AddExtraKeys.mm
```diff
@@ -0,0 +1,253 @@
...
+            case MOUSE_8:
+                return "Mouse 8";
+            default:
+                return CCKeyboardDispatcher::keyToString(KEY_Unknown);
+        }
+    }
+};
+
+
+#define OBJC_SWIZZLE(klass, methodName) \
```
qimiko, line 240: my new concern with swizzling (that i just made up) is that i think geode doing this would mess with other mods that also swizzle those methods - it might be good to make it a manual hook instead

ideally those mods would also switch to hooks but that requires i do something so... at least we can avoid making the problem worse lol

qimiko (changes requested):
it didn't crash so i'm otherwise calling it good

loader/src/hooks/AddExtraKeys.mm
SpaghettDev: I hook these methods in BetterInputs and it mostly works fine, but sure, I could change to manual hooks instead of swizzling

SpaghettDev:
> you forgot the android inputs in [`loader/src/platform/android/input.cpp`](https://github.com/geode-sdk/geode/blob/main/loader/src/platform/android/input.cpp) :(

i blame flame imo

qimiko (approved):
doesn't work on my iphone 17 air :frowning:

SpaghettDev:
> doesn't work on my iphone 17 air 😦

sorry, the iphone SE support is on its way..

matcool (approved):

PR #1266, opened 2025-03-12
Every custom constructor

hiimjasmine00:
In case y'all need them at some point


loader/include/Geode/cocos/robtop/content/CCContentManager.h
```diff
@@ -12,7 +12,7 @@ class CC_DLL CCContentManager : public cocos2d::CCObject
 {
 public:
     static CCContentManager* sharedManager();
-    CCContentManager();
+    CCContentManager() {}
```
matcool, line 15: wheres the macro?

PR #1279, opened 2025-04-02
Add enabling/disabling TextInput callback

SpaghettDev:


loader/include/Geode/ui/TextInput.hpp
```diff
@@ -43,6 +43,7 @@ namespace geode {
         CCTextInputNode* m_input;
         std::function<void(std::string const&)> m_onInput = nullptr;
         cocos2d::CCLabelBMFont* m_label = nullptr;
+        bool m_callbackEnabled = true;
```
matcool, line 46: abi break </3

Fleeym: see y'all in v5 :)

SpaghettDev:
why is ts a draft now

matcool:
it breaks abi, i wanted to mark it not ready so someone wouldnt accidentally merge it

loader/include/Geode/ui/TextInput.hpp
altalk23: why is this not impl

Fleeym:
hi we can break abi now

dankmeme01:
well we definitely can now!

SpaghettDev:
only took 10 months 🙏🙏🙏

PR #1282, opened 2025-04-04
Fix iOS FMOD

hiimjasmine00:
_My Lord... Is that... legal?_


matcool:
maybe make a macro to avoid all the repetition

hiimjasmine00:
> i request that you add assert test for gd version

I did that. I guess you couldn't find that in the sea of hooks.

FireMario211:
> > i request that you add assert test for gd version
> 
> I did that. I guess you couldn't find that in the sea of hooks.

I did not see that, never mind!

dankmeme01:
$on_mod(Loaded) instead of $execute perhaps? I don't know on what thread it runs on ios, but creating hooks on a thread could be problematic

dankmeme01 (approved):

PR #1295, opened 2025-04-19
make geode::utils::web::openLinkInBrowser open webview on android

Prevter:
Opens a browser in the launcher activity, which would prevent the game from closing.
Useful for cases where mod needs user to do something on a webpage and then return back (e.g. OAuth with local webserver)

Requires https://github.com/geode-sdk/android-launcher/pull/37

### DIFF
--- a/loader/include/Geode/utils/web.hpp
+++ b/loader/include/Geode/utils/web.hpp
@@ -9,15 +9,7 @@
 
 namespace geode::utils::web {
     GEODE_DLL void openLinkInBrowser(std::string const& url);
-
-#ifdef GEODE_IS_ANDROID
-    /**
-     * Opens a native webview window in the launcher activity.
-     * @param url URL to open.
-     */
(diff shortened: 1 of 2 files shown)


Prevter:
> Wouldn't it be better if we'd just add this functionality to geode::utils::web::openLinkInBrowser?

you do get an option to open the website in your browser when using CustomTabs, so could be an altervative i guess. might be even better since there's no need to add another method (and bump target geode version to use it).

Fleeym:
This also seems to be how most android apps handle opening urls

Prevter:
Alright, i'll change the existing method then

Fleeym:
Also, maybe on failure it could fallback to the old `CCApplication::sharedApplication()->openURL(url.c_str());`, in case the user has an older version of the launcher or something.

Fleeym:
Looks good to me, we can probably merge this without the android PR too :)

PR #1299, opened 2025-04-20
Remove Modtober 2024 tab from ModsLayer

Fleeym:
It's nearly May, we should probably hide this by now.

We could maybe reintroduce it in as an archive list, but the website has that anyway.


STGamer24YT:
Yeah I find it weird that this tab has been in Geode for **_months_**. I think it's time to remove it too.

PR #1304, opened 2025-04-22
macOS Shutdown Fix

hiimjasmine00:
This should make it build.


Fleeym:
Damn, alk left you hanging

PR #1313, opened 2025-04-28
Fix geode::TextInput not handling scale properly for touch

Alphalaneous:
This converts the touch to it's position relative to the parent scale if CCTextInputNode is within geode::TextInput. This resolves the issue where tapping/clicking on the input to change cursor pos is inaccurate.


dankmeme01 (approved):

Fleeym:
Tested, works for me too.

PR #1316, opened 2025-04-28
add format_as for seedvalues

Prevter:
> I wanted to add one but cvolton said they wanted too 
so i never did it
as you can see it's been a year and still no formatter 

> maybe one day a 3rd person will come in and actually add it


Fleeym:
vote peter for lead dev

PR #1320, opened 2025-04-30
Lazy loaded sprite

dankmeme01:
Adds a new class `geode::LazySprite`, which has APIs for loading an image from the internet, from a file, or from binary data.

Performs as much processing as possible in a background thread, to avoid freezes, and uses a lightweight threadpool, hence it's recommended that all mod developers who manually create images should use this class instead.

TODO: more docstrings and a docs page

### DIFF
--- a/loader/include/Geode/ui/LazySprite.hpp
+++ b/loader/include/Geode/ui/LazySprite.hpp
@@ -2,6 +2,7 @@
 
 #include "LoadingSpinner.hpp"
 #include <Geode/utils/web.hpp>
+#include <Geode/utils/cocos.hpp>
 
 #include <cocos2d.h>
 #include <filesystem>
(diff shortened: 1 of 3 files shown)


loader/include/Geode/ui/LazySprite.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+    protected:
+        Ref<LoadingSpinner> m_loadingCircle;
+        Callback m_callback;
+        Format m_expectedFormat;
+        EventListener<utils::web::WebTask> m_listener;
+        bool m_isLoading = false;
+        bool m_hasLoaded = false;
+        cocos2d::CCSize m_targetSize;
+        bool m_autoresize;
```
matcool, line 64: maybe make this private and/or make the class final? dont wanna risk abi breaks

PR #1321, opened 2025-04-30
Switch almost all AxisLayouts to SimpleAxisLayouts

Alphalaneous:
- Switch most AxisLayouts to SimpleAxisLayouts, excluding a few that may need a better alternative
- Clean up some node positioning and sizing
- Use only short modtober badge to remedy some scaling issues. 

These changes also slightly improve the performance of the UI, as AxisLayouts are expensive when updating.


hiimjasmine00:
This is so Fleeyms

Fleeym:
This PR doesn't build without PCH enabled, you should probably fix that.

PR #1324, opened 2025-05-03
Fix keyUp when command is down

Kingminer7:
Basically theres an issue where keyUp isn't called if command is down - this fixes that.

Only tested on intel, testing needed on m1

side note: not sure if this is 100% the way I should be hooking objc, lmk if its not


altalk23:
i can merge if you replace it with a ObjcHook::create hook

altalk23:
i guess this is a later me problem

PR #1330, opened 2025-05-06
Fix gd::vector and allow Android builds without pre-compiled headers

hiimjasmine00:


loader/src/platform/android/util.cpp
```diff
@@ -1,5 +1,3 @@
-using namespace geode::prelude;
```
dankmeme01, line 1: This is quite awesome

PR #1345, opened 2025-05-16
multipart/form-data support

Prevter:


loader/include/Geode/utils/web.hpp
```diff
@@ -60,6 +60,52 @@ namespace geode::utils::web {
...
+    public:
+        MultipartForm();
+
+        MultipartForm& param(std::string_view name, std::string_view value);
+        template <std::integral T>
+        MultipartForm& param(std::string_view name, T value) {
+            return this->param(name, std::to_string(value));
+        }
+
+        MultipartForm& file(std::string_view name, ByteVector const& data, std::string_view filename, std::string_view mime = "application/octet-stream");
```
matcool, line 78: `ByteVector const& data` since this cant move the vector, might as well use a `std::span<u8 const>`

loader/include/Geode/utils/web.hpp
```diff
@@ -60,6 +61,52 @@ namespace geode::utils::web {
...
+    private:
+        class Impl;
+
+        std::shared_ptr<Impl> m_impl;
+
+    public:
+        MultipartForm();
+
+        MultipartForm& param(std::string_view name, std::string_view value);
+        template <std::integral T>
```
Fleeym, line 74: Would it be worth it to also allow floats here?

Prevter: i copied this from existing WebRequest, so would make sense to also change that there

Fleeym: Out of the scope of the PR then, we'll do it in another one.

loader/src/utils/web.cpp
```diff
@@ -118,6 +119,138 @@ Result<> WebResponse::Impl::into(std::filesystem::path const& path) const {
...
+
+    void build() {
+        if (m_built) {
+            return;
+        }
+
+        this->generateBoundary();
+        m_data.clear();
+
+        const auto add_text = [&](std::string_view value) {
```
matcool, line 175: camel

loader/src/utils/web.cpp
```diff
@@ -118,6 +119,134 @@ Result<> WebResponse::Impl::into(std::filesystem::path const& path) const {
...
+    }
+    return *this;
+}
+
+Result<MultipartForm&> MultipartForm::file(std::string_view name, std::filesystem::path const& path, std::string_view mime) {
+    if (!m_impl->isBuilt()) {
+        GEODE_UNWRAP_INTO(auto data, utils::file::readBinary(path));
+        m_impl->m_files.insert_or_assign(std::string(name), MultipartFile{
+            .data = std::move(data),
+            .filename = path.filename().string(),
```
dankmeme01, line 219: wouldn't this be unsafe on windows? maybe get the wstring and convert it to utf8

PR #1350, opened 2025-05-18
Destructor fix

altalk23:
Replaces the CCDestructor destructor fix implementation with a better lock system that does not leak memory and does not crash on static destruction


ThePickleGawd:
```

PR #1375, opened 2025-06-03
linux installer fixes

RoootTheFox:
- fixes installer not checking for jq (but still depending on it)
- fixes invalid internal return codes (they're unsigned integers)
- improves syntax of checking if a command succeeded


RoootTheFox:
github actions out here building binaries for code that did not change :fire:

PR #1386, opened 2025-06-11
Read Description

stetupdev:
This pull request adds initial Linux platform detection and configuration support to Geode’s CMake build scripts.

PlatformDetect.cmake now detects Linux systems and sets GEODE_TARGET_PLATFORM accordingly.

Platform.cmake includes handling for Linux platforms to set target properties and linking placeholders.

This is an early implementation to enable building Geode on Linux. The changes have not yet been fully verified and will require testing and possible adjustments.

I plan to test this by cloning my fork with these changes and building on a Linux environment to confirm functionality.

### DIFF
--- a/cmake/Platform.cmake
+++ b/cmake/Platform.cmake
@@ -1,200 +1,219 @@
 include(cmake/PlatformDetect.cmake)
 
 if (NOT ${PROJECT_NAME} STREQUAL ${CMAKE_PROJECT_NAME})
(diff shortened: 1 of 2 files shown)


Cvolton:
May I ask what is the goal here? Geometry Dash doesn't have a native Linux version and you can already build the Windows version of Geode on Linux as is (in fact the current stable binaries are built on Linux to begin with)

PR #1389, opened 2025-06-14
Add API Tag

OmgRod:


altalk23:
can you integrate this with metadata.isAPI() as well

PR #1396, opened 2025-06-22
Android internal callbacks

qimiko:
good for future events
```c++
#include <Geode/Geode.hpp>
#include <Geode/loader/AndroidEvent.hpp>

$execute {
	auto launcherVersion = geode::utils::getLauncherVersion();
	if (launcherVersion.unwrapOr(0) < 23) {
		geode::log::warn("event timestamp method is unsupported by this device :(");
	}

	new geode::EventListener([](auto event) {
		auto timestamp = event->getTimestamp();
		geode::log::info("recv timestamp", timestamp);

		return geode::ListenerResult::Propagate;
	}, geode::AndroidInputTimestampFilter());
}
```


matcool:
loader version 23 should probably be mentioned in the docs for the input event

PR #1412, opened 2025-07-03
Added option to force curl to use IPv4

AlphiiGD:
I was having trouble with Curl. The literal only way I could get it to work was forcing it to use IPv4 instead of IPv6. Geode doesn't have an option to force IPv4 (not that I know of at least) so I added this. Made the shit work for me. Hopefully this could help someone in the future, too.

### DIFF
--- a/loader/src/utils/web.cpp
+++ b/loader/src/utils/web.cpp
@@ -562,6 +562,13 @@ WebTask WebRequest::send(std::string_view method, std::string_view url) {
         // Do not fail if response code is 4XX or 5XX
         curl_easy_setopt(curl, CURLOPT_FAILONERROR, 0L);
 
+		// IPv4
+		if (Mod::get()->getSettingValue<bool>("curl-force-ipv4")) {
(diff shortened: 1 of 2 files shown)


loader/src/utils/web.cpp
```diff
@@ -562,6 +562,13 @@ WebTask WebRequest::send(std::string_view method, std::string_view url) {
         // Do not fail if response code is 4XX or 5XX
         curl_easy_setopt(curl, CURLOPT_FAILONERROR, 0L);
 
+        // IPv4
+        if (Mod::get()->getSettingValue<bool>("curl-force-ipv4")) {
+            curl_easy_setopt(curl, CURLOPT_IPRESOLVE, CURL_IPRESOLVE_V4);
+        } else {
+            curl_easy_setopt(curl, CURLOPT_IPRESOLVE, CURL_IPRESOLVE_V6);
```
Fleeym: The default should be CURL_IPRESOLVE_WHATEVER imo

Fleeym (changes requested):

PR #1467, opened 2025-08-16
Progress bar helper classes

BlueWitherer:
Since the base game doesn't use a class to create its progress bar UI, I would like to propose some helper classes that I wrote that can help developers easily re-create them without having to use the sprites and set the texture rects themselves.

### `ProgressBar`
This class creates a progress bar that mimics the style of the progress bar shown when progressing through a classic level. It inherits `cocos2d::CCNode`.
```hpp
static ProgressBar* create(); // Create the custom progress bar

void setProgressBarFillColor(ccColor3B color); // Set the color of the fill of the bar
void updateProgress(float value); // Update the size of the fill of the bar

float getProgress(); // Get the current progress percentage of the bar
```

###### Example
```cpp
auto bar = ProgressBar::create();
bar->setProgressBarFillColor({255, 100, 100});
bar->updateProgress(20.5f); // Sets to 20.5% full
this->addChild(bar);
```

### `ProgressBarSolid`
This class, which inherits `ProgressBar`,  creates a progress bar that imitates the style of the Normal mode and Practice mode progress bars seen when pausing and before playing a level.
```hpp
static ProgressBarSolid* create(); // Create a custom progress bar

void updateProgress(float value) override; // Update the size of the fill of the bar

CCLabelBMFont* getProgressLabel(); // Get the progress percentage text label node
```

###### Example
```cpp
auto bar = ProgressBarSolid::create();
bar->setProgressBarFillColor({255, 100, 100});
bar->updateProgress(75.0f); // Sets to 75% full
this->addChild(bar);

auto barPercentLabel = bar->getProgressLabel();
barPercentLabel->setVisible(false);
```

### DIFF
--- a/loader/include/Geode/ui/ProgressBar.hpp
+++ b/loader/include/Geode/ui/ProgressBar.hpp
@@ -1,0 +1,42 @@
+#pragma once
+
+#include <cocos2d.h>
+
+namespace geode {
+    // Custom class for the progress bar
+    class ProgressBar : public cocos2d::CCNode {
+    protected:
+        Ref<CCSprite> m_progressBar = nullptr; // Progress bar outline
+        CCSprite* m_progressBarFill = nullptr; // Progress bar fill
+
+        float m_progress = 0.f; // Current progress bar fill percentage ranging from 0 to 100
+
+        float m_progressBarFillMaxWidth = 0.f; // Max width for the progress fill bar node
+        float m_progressBarFillMaxHeight = 0.f; // Max height for the progress fill bar node
+
+        bool init() override;
+
+    public:
+        // Create a custom progress bar
+        static ProgressBar* create();
+
+        /**
+         * Set the color of the fill of the bar
+         *
+         * @param color RGB color object
+         */
+        void setProgressBarFillColor(ccColor3B color);
+
+        /**
+         * Update the size of the fill of the bar
+         *
+         * @param value A float from 0 to 100
+         */
+        virtual void updateProgress(float value);
+
+        /**
+         * Get the current progress percentage of the bar
+         */
+        float getProgress();
+    };
+}
--- a/loader/include/Geode/ui/ProgressBarSolid.hpp
+++ b/loader/include/Geode/ui/ProgressBarSolid.hpp
@@ -1,0 +1,30 @@
+#pragma once
+
+#include <cocos2d.h>
+#include <Geode/ui/ProgressBar.hpp>
+
+namespace geode {
+    // Custom class for the Normal/Practice mode style progress bar
+    class ProgressBarSolid : public ProgressBar {
+    protected:
+        CCLabelBMFont* m_progressPercentLabel = nullptr; // The text label displaying the percentage
+
+        bool init() override;
+
+    public:
+        // Create a custom progress bar
+        static ProgressBarSolid* create();
+
+        /**
+         * Update the size of the fill of the bar and the percentage label
+         *
+         * @param value A float from 0 to 100
+         */
+        void updateProgress(float value) override;
+
+        /**
+         * Get the progress percentage text label node
+         */
+        CCLabelBMFont* getProgressLabel();
+    };
+}
--- a/loader/src/ui/nodes/ProgressBar.cpp
+++ b/loader/src/ui/nodes/ProgressBar.cpp
@@ -1,0 +1,67 @@
+#include <cocos2d.h>
+#include <Geode/ui/ProgressBar.hpp>
+
+using namespace geode::prelude;
+
+bool ProgressBar::init() {
+    if (!CCNode::init()) return false;
+
+    this->setID("bar"_spr);
+
+    m_progressBar = CCSprite::create("slidergroove2.png");
+    m_progressBar->setID("progress-bar");
+    m_progressBar->setPosition({ m_progressBar->getScaledContentWidth() / 2.f, m_progressBar->getScaledContentHeight() / 2.f });
+    m_progressBar->setAnchorPoint({ 0.5, 0.5 });
+    m_progressBar->setZOrder(501);
+
+    this->setScaledContentSize(m_progressBar->getScaledContentSize());
+
+    m_progressBarFill = CCSprite::create("sliderBar2.png");
+    m_progressBarFill->setID("bar-fill");
+    m_progressBarFill->setAnchorPoint({ 0, 0.5 });
+    m_progressBarFill->setPosition({ 2.f, m_progressBar->getScaledContentHeight() / 2.f });
+    m_progressBarFill->setColor({ 255, 255, 255 });
+    m_progressBarFill->setZOrder(-1);
+
+    m_progressBarFillMaxWidth = m_progressBar->getScaledContentWidth() - 4.f;
+    m_progressBarFillMaxHeight = m_progressBarFill->getScaledContentHeight();
+
+    m_progressBar->addChild(m_progressBarFill);
+
+    this->addChild(m_progressBar);
+    this->updateProgress(0.0f);
+
+    return true;
+};
+
+void ProgressBar::setProgressBarFillColor(ccColor3B color) {
+    if (m_progressBarFill) m_progressBarFill->setColor(color);
+};
+
+void ProgressBar::updateProgress(float value) {
+    if (value > 100.0f) value = 100.0f;
+    if (value < 0.0f) value = 0.0f;
+
+    m_progress = value;
+
+    if (m_progressBar && m_progressBarFill) {
+        float width = m_progressBarFillMaxWidth * (m_progress / 100.f);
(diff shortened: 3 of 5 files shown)


loader/include/Geode/ui/ProgressBarSolid.hpp
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include <cocos2d.h>
+#include <Geode/ui/ProgressBar.hpp>
+
+namespace geode {
+    // Custom class for the Normal/Practice mode style progress bar
+    class ProgressBarSolid : public ProgressBar {
```
HJfod: Is there a reason for having this be a separate class over just a style parameter on the normal `ProgressBar` class? The only member this adds is the progress bar percentage, which sounds like it would be useful as an option for any progress bar

loader/include/Geode/ui/ProgressBar.hpp
```diff
@@ -0,0 +1,42 @@
...
+         * Update the size of the fill of the bar
+         *
+         * @param value A float from 0 to 100
+         */
+        virtual void updateProgress(float value);
+
+        /**
+         * Get the current progress percentage of the bar
+         */
+        float getProgress();
```
HJfod: Getters should always be marked `const`

loader/include/Geode/ui/ProgressBar.hpp
```diff
@@ -0,0 +1,42 @@
...
+    // Custom class for the progress bar
+    class ProgressBar : public cocos2d::CCNode {
+    protected:
+        Ref<CCSprite> m_progressBar = nullptr; // Progress bar outline
+        CCSprite* m_progressBarFill = nullptr; // Progress bar fill
+
+        float m_progress = 0.f; // Current progress bar fill percentage ranging from 0 to 100
+
+        float m_progressBarFillMaxWidth = 0.f; // Max width for the progress fill bar node
+        float m_progressBarFillMaxHeight = 0.f; // Max height for the progress fill bar node
```
HJfod: Class members should be PIMPL; aka these should be behind a source-defined `Impl` class and `std::unique_ptr<Impl> m_impl`. See other Geode classes for examples

loader/src/ui/nodes/ProgressBar.cpp
```diff
@@ -0,0 +1,67 @@
...
+bool ProgressBar::init() {
+    if (!CCNode::init()) return false;
+
+    this->setID("bar"_spr);
+
+    m_progressBar = CCSprite::create("slidergroove2.png");
+    m_progressBar->setID("progress-bar");
+    m_progressBar->setPosition({ m_progressBar->getScaledContentWidth() / 2.f, m_progressBar->getScaledContentHeight() / 2.f });
+    m_progressBar->setAnchorPoint({ 0.5, 0.5 });
+    m_progressBar->setZOrder(501);
```
HJfod: Is there a reason for this? Seems like an arbitary magic number

HJfod (changes requested):
Thank you very much for this PR! Left a few comments & fix suggestions, but looking good overall :)

BlueWitherer:
appreciate the feedback! i've made adjustments based on what you mentioned and asked for - lmk if anything else could use further tweaking :D

loader/include/Geode/ui/ProgressBar.hpp
```diff
@@ -0,0 +1,70 @@
...
+         * @param style Style to switch to
+         */
+        void setStyle(ProgressBarStyle style);
+
+        /**
+         * Set the color of the fill of the bar
+         *
+         * @param color RGB color object
+         */
+        void setProgressBarFillColor(ccColor3B color);
```
matcool: personally just `setFillColor` is enough

loader/include/Geode/ui/ProgressBar.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+        /**
+         * Get the progress percentage text label node
+         */
+        CCLabelBMFont* getProgressLabel() const;
+
+        /**
+         * Get the current style of the progress bar
+         */
+        ProgressBarStyle getProgressBarStyle() const;
```
matcool: should just be `getStyle`, like how `setStyle` is

loader/src/ui/nodes/ProgressBar.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+    float progress = 0.0f; // Current progress bar fill percentage ranging from 0 to 100
+
+    ccColor3B progressBarFillColor = { 255, 255, 255 }; // Current color of the filled progress bar
+    bool showProgressPercentLabel = false; // Whether to show the label showing the percentage of the current progress
+
+    ProgressBarStyle style = ProgressBarStyle::Level; // Style of the progress bar
+
+    float progressBarFillMaxWidth = 0.0f; // Max width for the progress fill bar node
+    float progressBarFillMaxHeight = 0.0f; // Max height for the progress fill bar node
```
matcool: imo comments should be on the line before, not at the end of the line

loader/src/ui/nodes/ProgressBar.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+    return true;
+};
+
+void ProgressBar::setStyle(ProgressBarStyle style) {
+    if (m_impl->style != style) {
+        m_impl->style = style;
+
+        this->removeAllChildren();
+        init(); // init again with new style
```
matcool: calling CCNode::init multiple times 😟

BlueWitherer:
sorry about all that, hopefully this fixes it!

loader/include/Geode/ui/ProgressBar.hpp
```diff
@@ -0,0 +1,77 @@
...
+    };
+
+    // Custom class for the progress bar
+    class ProgressBar : public cocos2d::CCNode {
+    protected:
+        class Impl;
+        std::unique_ptr<Impl> m_impl;
+
+        ProgressBar();
+        ~ProgressBar();
```
HJfod: The destructor should be marked `virtual`

loader/src/ui/nodes/ProgressBar.cpp
```diff
@@ -0,0 +1,178 @@
...
+        m_impl->progressBarFillMaxHeight = m_impl->progressBarFill->getScaledContentHeight() - 0.5f;
+
+        m_impl->progressPercentLabel = CCLabelBMFont::create("0%", "bigFont.fnt");
+        m_impl->progressPercentLabel->setID("progress-percent-label");
+        m_impl->progressPercentLabel->setScale(0.5f);
+        m_impl->progressPercentLabel->setAnchorPoint({ 0, 0.5 });
+        m_impl->progressPercentLabel->setPosition({ m_impl->progressBar->getScaledContentWidth() + 2.5f, m_impl->progressBar->getScaledContentHeight() / 2.0f });
+        m_impl->progressPercentLabel->setAlignment(CCTextAlignment::kCCTextAlignmentLeft);
+        m_impl->progressPercentLabel->setVisible(m_impl->showProgressPercentLabel);
+        m_impl->progressPercentLabel->setZOrder(1);
```
HJfod: This is just a styling thing but I would indent this whole thing one tab and then surround the block with braces, like so:
```cpp
switch (x) {
    case y: {
        // code here...
    } break;
}
```

loader/include/Geode/ui/ProgressBar.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+        void customSetup();
```
HJfod: I tend to call functions like this `reload` or `reloadStyle` instead to make it clear that it may (and should) be called multiple times

loader/src/ui/nodes/ProgressBar.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+    m_impl->progress = value;
+
+    if (m_impl->progressBarFill) {
+        float width = m_impl->progressBarFillMaxWidth * (m_impl->progress / 100.0f);
+        m_impl->progressBarFill->setTextureRect({ 0.0f, 0.0f, width, m_impl->progressBarFillMaxHeight });
+    };
+
+    if (m_impl->progressPercentLabel) {
+        auto percentString = fmt::format("{}%", static_cast<int>(m_impl->progress));
```
HJfod: This should probably use `geode::utils::numToString` with a precision of 0. The class should probably also have a way to customize the precision of the percentage

HJfod (approved):
Looks good to me, don't have any major complaints, just some styling suggestions! Tyvm for this PR!

BlueWitherer:
happy to contribute! hopefully i've made this as best as it could possibly be!

HJfod (approved):
LGTM!

BlueWitherer:
happy to contribute!

PR #1470, opened 2025-08-24
(v5) Overlay system

Kingminer7:
Most (if not all) mods that use keepAcrossScenes try to keep their node above the current scene, so why not just have an overlay to make that easier? I've added an overlay that gets visited after the CCScene, so it'll always render above and won't have weirdness moving the node between changing CCScenes.

I also have a [fork of DevTools](https://github.com/Kingminer7/DevTools/tree/v5-overlay) prepared which renders the overlay node below the CCScene.

Marked as draft as it contains breaking changes and should only be merged with geode v5.

Credit to Alphalaneous for the idea on how to implement this!


loader/src/hooks/persist.cpp
```diff
@@ -11,26 +12,19 @@ using namespace geode::prelude;
...
 
-#ifdef GEODE_IS_WINDOWS
-struct SceneSwitch : Modify<SceneSwitch, AppDelegate> {
+struct DrawOverlay : Modify<DrawOverlay, CCEGLView> {
     GEODE_FORWARD_COMPAT_DISABLE_HOOKS("persist disabled")
-    void willSwitchToScene(CCScene* scene) {
-        AppDelegate::willSwitchToScene(scene);
-        SceneManager::get()->willSwitchToScene(scene);
+    static void onModify(auto& self) {
+        if (auto res = self.setHookPriorityPost("cocos2d::CCLabelBMFont::init", Priority::First); !res) {
```
matcool: ???

Kingminer7: oh whoops that was a mistake

Fleeym:
You should target this against the v5 branch and remove the draft if you think it's done

Kingminer7:
Done, also feel free to change the name of it since I'm not the greatest at naming

dankmeme01:
looks mostly good to me, curious why not use CCDirector notification node? Both approaches are probably nearly identical, but notification node is hookless and feels less hacky; but idk about all the upsides/downsides of doing so

Kingminer7:
alk just a hater frfr

altalk23:
ooks good to go except for the part that notification node doesnt take in input so you cant use stuff like buttons with it

altalk23:
its manually visited in drawscene

Alphalaneous:
True, tho that can be fixed by running onEnter on any node that is added and onExit when removed

Edit: Already calls onEnter on the node itself

PR #1481, opened 2025-09-03
Pass progress to mapped task in Task::chain

prevter:


matcool:
wouldnt this mean the progress jumps all over the place? like for example if you chain two download tasks, first one would send progress from 0 to 100%, and then the second one starts and resets it back to 0

prevter:
> wouldnt this mean the progress jumps all over the place? like for example if you chain two download tasks, first one would send progress from 0 to 100%, and then the second one starts and resets it back to 0

yeah, i guess that's one issue with this approach (not that i've seen anyone use this particular method and/or progress with it), maybe should've just copied the entire function and added a progress mapper like in the todo comment

PR #1501, opened 2025-09-30
Added Modtober 2025 content

SMJSGaming:
This was mostly based on the Modtober 2024 content with a few slight adjustments for permanent content to prevent a lot of duplicate code.

It's all tested.


loader/src/ui/mods/popups/ModPopup.cpp
```diff
@@ -1167,6 +1168,15 @@ void ModPopup::onModtoberInfo(CCObject*) {
     )->show();
 }
 
+void ModPopup::onModtober25Info(CCObject*) {
+    FLAlertLayer::create(
+        "Modtober 2025",
+        "This mod was an entry for <co>Modtober 2025</c>, a contest to create "
```
Fleeym: This should probably say "This mod *is* an entry for Modtober 2025"

SMJSProductions: Was debating it a bit due to it requiring a second update then but maybe ye

PR #1541, opened 2025-11-19
Added an "Installation" section to the README

xblazegmd:
I think we should maybe be clear about how to even get Geode in the first place lol


Lilicogamer13:
why didnt geode already simply have that 😭

PR #1542, opened 2025-11-22
Add Back Button function for layers

BlueWitherer:
A simple convenient function for developers to quickly add a very basic back button onto their custom UI layers. Here's an example of a pink back button applied to this template layer.
image

The function is defined as such.
```hpp
    void addBackButton(
        cocos2d::CCNode* to,
        std::function<void(cocos2d::CCMenuItem*)> callback,
        BackButtonStyle style = BackButtonStyle::Green
    );
```

It's as simple as pie!
```cpp
geode::addBackButton(this); // use default callback and style!
```
```cpp
geode::addBackButton(this, [](CCMenuItem*) { /* your logic here */ }, BackButtonStyle::Pink); // make it pretty!
```

It can get annoying for developers having to manually create this button every time especially when they're working with multiple custom layers; the intention behind this is to simply streamline that process.

originally suggested by @DumbCaveSpider :3

### DIFF
--- a/loader/include/Geode/ui/General.hpp
+++ b/loader/include/Geode/ui/General.hpp
@@ -105,9 +105,10 @@ namespace geode {
     };
 
     /**
-     * Add a back button to a layer, will call `CCDirector::get()->popScene()` on activate
-     * @param to Layer to add back button to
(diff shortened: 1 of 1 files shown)


BlueWitherer:
just learned gametoolbox has this but it just crashes when used

PR #1543, opened 2025-11-24
Fix user protocol doc oversight

Nightcaat:
The user protocol actually requires user ID, not account ID :3


Cvolton:
that doesn't seem to be right, since the `user:761691` in the Geode about.md resolves to my account where the account ID is 761691 and user ID is 6330800

and the protocol implementation just opens a ProfilePage directly in a single step, something you can't even do just using a user ID

PR #1556, opened 2025-12-12
Implement fallback for rendering CCClippingNode

FireMario211:
This implements a "fallback rendering" when stencil buffer is disabled.

This is a workaround for getting ANGLEGLKit to work with CCClippingNode, albeit in a limited state. The workaround being to use glScissor instead, as most mods that use CCClippingNode usually have the mask as a box or rectangle.

Also boundingBox is used as some mods dont properly set the anchor points, causing the mask to be in the wrong position.


FireMario211:
Before:
image

After:
image

PR #1572, opened 2025-12-23
remove support for mod.json gd as string

Fleeym:
Not sure if there's any other place where this is checked, will take a look into the code.


fabiangrondona09-svg:
Geode do let me download

PR #1592, opened 2026-01-15
Add modify friends, missing functions and custom constructors to cocos2d classes

hiimjasmine00:
Also a few minor changes

(Note: cocos2d::extension::CCControlColourPicker::create and cocos2d::extension::CCControlColourPicker::colourPicker should be swapped, but I don't want to break existing bindings)


appleTrackerTree (commented):
Cool

PR #1630, opened 2026-01-22
Add Input event API

prevter:
Since 2.2081 removed onGLFW callbacks, the only really viable way to get inputs is using the WndProc hook. Instead of having each mod hooking it, Geode will handle everything and provide events.
Events probably should work on all platforms (android, mac, ios), but I will leave the implementation details to someone who can test them (or try with a VM later).

Tested on Windows with Eclipse Menu

### DIFF
--- a/loader/src/loader/LoaderImpl.cpp
+++ b/loader/src/loader/LoaderImpl.cpp
@@ -404,7 +404,7 @@ void Loader::Impl::buildModGraph() {
                 continue;
             }
 
-            if (dependency.isRequired() || dependency.getMod() == nullptr) {
+            if (!dependency.isRequired() || dependency.getMod() == nullptr) {
                 continue;
             }
 
--- a/loader/src/loader/ModMetadataImpl.cpp
+++ b/loader/src/loader/ModMetadataImpl.cpp
@@ -853,7 +853,15 @@ ModMetadata ModMetadata::createFromGeodeFile(std::filesystem::path const& path) 
     }
     auto&& modJson = std::move(modJsonRes.unwrap());
 
-    return Impl::parse(modJson, guessedID);
+    auto info = Impl::parse(modJson, guessedID);
+    info.m_impl->m_path = path;
+
+    auto specialFilesRes = info.m_impl->addSpecialFiles(unzip);
+    if (!specialFilesRes) {
+        info.m_impl->m_errors.emplace_back(fmt::format("Unable to add extra files: {}", specialFilesRes.unwrapErr()));
+    }
+
+    return info;
 }
 ModMetadata ModMetadata::create(ModJson const& json) {
     return Impl::parse(json, std::nullopt);


iAndyHD3:
looks good, I would add mouse scroll wheel before controller inputs. not sure if CCMouseDispatcher works but even if it does it's probably better to go all in with system apis.

altalk23:
pumping inputs raw since 2026 (thank you robtop)

loader/include/Geode/utils/Keyboard.hpp
```diff
@@ -0,0 +1,98 @@
...
+        double timestamp;
+        cocos2d::enumKeyCodes key;
+        Action action;
+        Modifiers modifiers = Mods_None;
+
+        KeyboardInputData(cocos2d::enumKeyCodes key, Action action, Native native, double timestamp, Modifiers mods) noexcept
+            : native(native), timestamp(timestamp), key(key), action(action), modifiers(mods) {}
+    };
+
+    struct KeyboardInputEvent final : GlobalEvent<KeyboardInputEvent, bool(KeyboardInputData&), bool(KeyboardInputData&), cocos2d::enumKeyCodes> {
```
altalk23: this should be just GlobalEvent<KeyboardInputEvent, bool(KeyboardInputData&, cocos2d::enumKeyCodes>

loader/src/platform/ios/input.mm
```diff
@@ -0,0 +1,332 @@
...
+    );
+
+    auto res = KeyboardInputEvent(keyCode).send(data);
+
+    isDown = data.action != KeyboardInputData::Action::Release;
+    isRepeat = data.action == KeyboardInputData::Action::Repeat;
+    keyCode = data.key;
+    modifiers = data.modifiers;
+
+    // TODO: Find CCIMEDispatcher::sharedDispatcher and reimplement CCIMEDispatcher::hasDelegate
```
altalk23: im assuming this todo is done?

loader/include/Geode/utils/Keyboard.hpp
```diff
@@ -0,0 +1,98 @@
...
+        enum Modifiers : uint8_t {
+            Mods_None    = 0,
+
+            Mods_Shift   = 1 << 0,
+            Mods_Control = 1 << 1,
+            Mods_Alt     = 1 << 2,
+            Mods_Super   = 1 << 3,
+        };
+
+        struct Native final {
```
altalk23, line 22: possibly have the native as an opaque handle? and have getVKey and getScanCode implemented for it in platform separately?

altalk23 (changes requested):

altalk23:
guys 1 day ban from discord this is funny

loader/src/platform/ios/input.mm
hiimjasmine00: Yes, it should be

PR #1631, opened 2026-01-22
move m_ignoreInvisibleChildren out of Layout

Fleeym:
Also makes it default to true, which is a very cool and good change.

### DIFF
--- a/loader/include/Geode/ui/Layout.hpp
+++ b/loader/include/Geode/ui/Layout.hpp
@@ -307,7 +307,7 @@ public:
      * calculating content sizes
      */
     AxisLayout* ignoreInvisibleChildren(bool ignore = true);
-    bool isIgnoreInvisibleChildren();
+    bool isIgnoreInvisibleChildren() const;
 };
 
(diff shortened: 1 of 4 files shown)


SpaghettDev:
why did u move it out of the Layout class? seems kinda counter productive since all subclasses use the member :[]()sob:

matcool:
anchorlayout does not

SpaghettDev:
so duplicate code is better than a bool member not being used in 1 subclass?

BlueWitherer:
> anchorlayout does not

AnchorLayout::apply uses Layout::getNodesToPosition internally which also depended on m_ignoreInvisibleChildren being present within Layout

chagh-dev:
this PR broke geode 🥰
`Layout::ignoreInvisibleChildren` and `Layout::isIgnoreInvisibleChildren` implementations were not removed
`Layout::getNodesToPosition` is a non virtual function that implements "ignore invisible" functionality. So we either need to copy this as well or figure something out

PR #1638, opened 2026-01-23
WebRequest API rewrite

prevter:
- Using a single thread worker that manages all requests, instead of thread-per-request (curl_multi)
- Allow sync requests, to bypass queueInMainThread if you already have an existing thread
- Make WebProgress not pimpl to minimize overhead
- Added rvalue overload for data() in WebResponse to get the bytes without copies

### DIFF
--- a/loader/src/utils/web.cpp
+++ b/loader/src/utils/web.cpp
@@ -967,7 +967,7 @@ public:
                     requestData.response.m_impl->m_code = static_cast<int>(code);
 
                     char* errorBuf = requestData.request->m_errorBuf;
-                    requestData.response.m_impl->m_errMessage = errorBuf ? std::string(errorBuf) : "";
+                    requestData.response.m_impl->m_errMessage = std::string(errorBuf);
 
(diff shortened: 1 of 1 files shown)


dankmeme01:
i've said some of this on discord but putting here so it doesn't get forgotten:

1. Worker thread is busy looping when there are no pending requests, as `curl_multi_poll` will immediately return. Not entirely sure on the solution, but the thread could have an internal count of in-flight request and at the start of each loop iteration, block on some semaphore if that count is 0 (and functions that push requests release the semaphore). feel free to do any different way though
2. Allegedly it causes issues on Wine when a huge amount of requests are in-flight, but this seems to be a Wine bug so idk if you need to solve this
3. WebRequestManager is a static, and cleanup in statics is bug prone. This might be a non-issue here, but the curl cleanup in destructor can cause occasional crashes/hangs on Windows. In that case, it should be completely fine to leak the impl by doing `m_impl.release()` in the dtor

other than that, everything looks fantastic :)

PR #1648, opened 2026-01-26
Make macOS build after #1638

xblazegmd:
After the WebRequest changes in #1638, macOS now requires requires CMAKE_OSX_DEPLOYMENT_TARGET to be set to macOS 11.0 or later, instead of the usual 10.15. Otherwise it won't build.

Although there probably needs to be some changes since the game is pretty buggy after the change (game can freeze sometimes). I cannot fully tell what's causing the issue, whether the changes in how WebRequests work or 11.0, but if I ever find the cause I'll see if I can fix it (or at least try to).

If CMAKE_OSX_DEPLOYMENT_TARGET cannot be set to 11.0 for whatever reason (maybe for the previous reasons), there needs to be a change in how WebRequests work after #1638 since that uses `std::binary_semaphore`, which is only available on macOS 11.0 or later

### DIFF
--- a/loader/include/Geode/platform/platform.hpp
+++ b/loader/include/Geode/platform/platform.hpp
@@ -14,9 +14,15 @@
 #define GEODE_WRAPPER_STR(...) #__VA_ARGS__
 #define GEODE_STR(...) GEODE_WRAPPER_STR(__VA_ARGS__)
 
+#if defined (_MSC_VER) && !defined(__clang__)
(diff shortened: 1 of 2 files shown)


hiimjasmine00:
Look into cmake/Platform.cmake. That should be where the fix is placed.

hiimjasmine00:
The last change should be in loader/CMakeLists.txt on line 341.

PR #1668, opened 2026-01-31
notification rewrite

undefined06855:
also made loadingspinner a ccnodergba so that it can fade in/out

### DIFF
--- a/loader/include/Geode/ui/Notification.hpp
+++ b/loader/include/Geode/ui/Notification.hpp
@@ -33,11 +33,9 @@ namespace geode {
         void showNextNotification();
         void waitThenHide();
 
-    private:
         NineSlice* getBG();
         cocos2d::CCLabelBMFont* getLabel();
         cocos2d::CCNodeRGBA* getContent();
-        bool isShowing();
 
     public:
         /**
@@ -76,6 +74,7 @@ namespace geode {
         void setTime(float time);
 
         float getTime();
+        bool isShowing();
 
         /**
          * Adds the notification to the current scene if it doesn't have a
@@ -99,3 +98,4 @@ namespace geode {
         void cancel();
     };
 }
+


loader/src/ui/nodes/Notification.cpp
```diff
@@ -238,19 +235,36 @@ void Notification::wait() {
...
+
+        CCDelayTime::create(NOTIFICATION_FADEOUT * .75f),
+
         CCCallFunc::create(this, callfunc_selector(Notification::showNextNotification)),
         nullptr
     ));
 }
 
 void Notification::cancel() {
-    if(m_pParent) return this->hide();
```
altalk23, line 251: is it safe to remove the parent check?

loader/src/ui/nodes/Notification.cpp
```diff
@@ -136,98 +135,96 @@ void Notification::setIcon(NotificationIcon icon) {
...
     }
-    m_impl->bg->runAction(CCFadeTo::create(NOTIFICATION_FADEIN, 150));
-}
 
-void Notification::animateOut() {
-    m_impl->label->runAction(CCFadeTo::create(NOTIFICATION_FADEOUT, 0));
-    if (m_impl->icon) {
-        m_impl->icon->runAction(CCFadeTo::create(NOTIFICATION_FADEOUT, 0));
+    // if we're not the current notification, return
+    if (s_queue.at(0) != this) {
```
altalk23, line 189: is it safe to do indexing without size checking here?

loader/src/ui/nodes/Notification.cpp
```diff
@@ -5,15 +5,16 @@
 using namespace geode::prelude;
 
 constexpr auto NOTIFICATION_FADEIN = .3f;
-constexpr auto NOTIFICATION_FADEOUT = 1.f;
+constexpr auto NOTIFICATION_FADEOUT = .5f;
 
-CCArray* Notification::s_queue = nullptr;
+std::vector<Ref<Notification>> Notification::s_queue;
```
altalk23: would it be a better idea to move the queue to source entirely?

altalk23: also change it to a deque from vector

loader/src/ui/nodes/Notification.cpp
undefined06855: hide doesn't do anything with the parent (and unless you're doing wacky stuff there should be a parent if m_impl->showing is true)

i think fod just used m_pParent instead of a showing check for whatever reason?

loader/src/ui/nodes/Notification.cpp
undefined06855: > would it be a better idea to move the queue to source entirely?

i just left it as close to whatever fod wrote originally, but ill move it and change to deque

loader/src/ui/nodes/Notification.cpp
altalk23: okay it does seem like its safe as hide doesnt remove from parent directly, but showNextNotification does, so does that have any possibility of crashing?

loader/src/ui/nodes/Notification.cpp
undefined06855: if it's not in the queue, it'll be appended in the line above, and if the queue is empty it will always be appended in the line above (since it's not in the queue)

loader/src/ui/nodes/Notification.cpp
undefined06855: ccnode removeFromParent checks if parent is nullptr before doing anything with it

loader/include/Geode/ui/Notification.hpp
```diff
@@ -25,19 +22,22 @@ namespace geode {
...
+        static cocos2d::CCNode* createIcon(NotificationIcon icon);
 
-        void animateIn();
-        void animateOut();
         void showNextNotification();
-        void wait();
+        void waitThenHide();
+
+    private:
+        NineSlice* getBG();
```
altalk23, line 36: why are there private functions if the class is pimpl, these are prob better as protected funcs

altalk23 (changes requested):

loader/include/Geode/ui/Notification.hpp
undefined06855: oops

PR #1676, opened 2026-02-01
Templated ZStringView

hiimjasmine00:


hiimjasmine00:
I just realized this doesn't build at all so hold on

PR #1683, opened 2026-02-03
Changes to changelog

xblazegmd:
- Change some wording
- Add a "breaking change" indicator

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Geode Changelog
 
 ## v5.0.0-alpha.1 (WIP changelog)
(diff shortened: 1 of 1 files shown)


xblazegmd:
With "breaking changes" I mean anything that may require a rewrite of v4 mods. There's some stuff I was unsure abt marking as breaking so if there's something that should (or shouldn't) be marked as breaking lmk

matcool:
idk the breaking changes being in changelog.md is kinda noisy, i dont like it

xblazegmd:
> idk the breaking changes being in changelog.md is kinda noisy, i dont like it

@matcool Well I thought it would be helpful but yeah, now that I think about it maybe not.

I'll try and remove them, and maybe make a new PR with just the wording changes if you're fine with it (although idk what you may think of those)

PR #1695, opened 2026-02-05
Improve ProgressBar Impl

BlueWitherer:
I've proposed the following breaking changes to `geode::ProgressBar`'s implementation.
- Added **`noexcept`** to getters
- Added a `geode::ProgressBarStyle` optional parameter in **`ProgressBar::create`**

Among other small internal tweaks.

### DIFF
--- a/loader/include/Geode/ui/ProgressBar.hpp
+++ b/loader/include/Geode/ui/ProgressBar.hpp
@@ -18,20 +18,7 @@ namespace geode {
         ProgressBar();
         ~ProgressBar();
 
-        /**
-         * Set up `Level`/`Slider` progress bar style
-         *
-         * @note This is mainly to avoid repeating code since the only difference between these is two sprite names
-         *
-         * @param barSpr Sprite of the bar
-         * @param fillSpr Sprite of the fill
-         */
-        void setupBarStyle(const char* barSpr, const char* fillSpr);
-
-        // Set up `Solid` progress bar style
-        void setupSolidStyle();
-
-        // Reloads the style of the progress bar, calls either `setupBarStyle` or `setupSolidStyle` internally depending on style
+        // Reloads the style of the progress bar
         void reloadStyle();
 
         bool init(ProgressBarStyle style);
--- a/loader/src/ui/nodes/ProgressBar.cpp
+++ b/loader/src/ui/nodes/ProgressBar.cpp
@@ -17,7 +17,7 @@ public:
 
     // Style of the progress bar
     ProgressBarStyle style = ProgressBarStyle::Level;
-    // Current color of the filled progress bar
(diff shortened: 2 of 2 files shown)


loader/src/ui/nodes/ProgressBar.cpp
```diff
@@ -97,7 +97,7 @@ void ProgressBar::reloadStyle() {
     } break;
     };
 
-    this->setScaledContentSize(m_impl->progressBar->getScaledContentSize());
```
altalk23, line 100: is this change fine? id assume setcontentsize would need getcontentsize instead of the scaled one, but im not entirely sure how it works

BlueWitherer: ah, made a little oversight here since this assumes every time it reloads the node is always at scale 1, will fix!

loader/src/ui/nodes/ProgressBar.cpp
```diff
@@ -126,6 +129,25 @@ void ProgressBar::setStyle(ProgressBarStyle style) {
     };
 };
 
+void ProgressBar::setGoldVariant(bool enable) {
+    if (m_impl->isGold != enable) {
+        m_impl->isGold = enable;
+
+        if (m_impl->style == ProgressBarStyle::Level) {
```
altalk23: would be better to add the gold variant to the enum instead if it only affects the level type

altalk23 (changes requested):

loader/src/ui/nodes/ProgressBar.cpp
BlueWitherer: i can certainly try that! i originally went with the boolean since it would only need to change the outline sprite for the level type

loader/include/Geode/ui/ProgressBar.hpp
```diff
@@ -4,28 +4,45 @@
...
-        // Reloads the style of the progress bar
+        /**
+         * Set up `Level`/`Slider` progress bar style
+         *
+         * @note This is mainly to avoid repeating code since the only difference between these is two sprite names
+         *
+         * @param barSpr Sprite of the bar
+         * @param fillSpr Sprite of the fill
+         */
+        void setupBarStyle(const char* barSpr, const char* fillSpr);
```
altalk23: is it fine to have these as protected istead of just in impl

altalk23:
otherwise lgtm

loader/include/Geode/ui/ProgressBar.hpp
BlueWitherer: yeah lol actually it would be better to put these in the pimpl, i'll get that done soon!

PR #1696, opened 2026-02-05
Add SceneEvent

Alphalaneous:
Very simple event for when you switch scenes. A lot of mods hook willSwitchToScene, so I felt this would be useful.


matcool:
can u add some /// docs to the class? like specifying its triggered before the switch actually happens, etc

Alphalaneous:
> can u add some /// docs to the class? like specifying its triggered before the switch actually happens, etc

Done

matcool:
lol sure

PR #1697, opened 2026-02-05
Web intercepting API

SMJSProductions:


loader/include/Geode/utils/web.hpp
```diff
@@ -129,6 +130,11 @@ namespace geode::utils::web {
 
     class GEODE_DLL WebResponse final {
     private:
+        using Listener = Function<void(const WebResponse&)>;
+
+        static std::unordered_map<std::string, std::vector<std::shared_ptr<Listener>>> s_listeners;
+        static std::shared_mutex s_listenersMutex;
```
dankmeme01: I find that having them be handled by WebResponse/WebRequest is rather weird, makes a lot more sense to have these static variables privately defined in the .cpp file and register functions to just be in the `geode::utils::web` namespace

Also, use `StringMap` from `Geode/utils/StringMap.hpp`

(this comment applies entirely to both listeners and interceptors)

loader/src/utils/web.cpp
```diff
@@ -437,7 +490,7 @@ class WebRequest::Impl {
         auto res = WebResponse();
         res.m_impl->m_code = static_cast<int>(code);
         res.m_impl->m_data = ByteVector(msg.begin(), msg.end());
-        return res;
+        return std::move(res);
```
dankmeme01: this move doesn't add anything and only hurts an existing optimization (NRVO)

loader/src/utils/web.cpp
```diff
@@ -734,40 +790,70 @@ std::atomic_size_t WebRequest::Impl::s_idCounter = 0;
 WebRequest::WebRequest() : m_impl(std::make_shared<Impl>()) {}
 WebRequest::~WebRequest() {}
 
-WebFuture WebRequest::send(std::string method, std::string url) {
+WebFuture WebRequest::send(std::string method, std::string url, Mod* mod) {
+    if (m_impl->m_inInterceptor) throw std::runtime_error("Don't send a request again from an interceptor.");
```
dankmeme01: i think this could be worded a bit less directly, like "Cannot call send again while inside an interceptor callback"

loader/src/utils/web.cpp
```diff
@@ -416,6 +467,8 @@ class WebRequest::Impl {
     ProxyOpts m_proxyOpts = {};
     HttpVersion m_httpVersion = HttpVersion::DEFAULT;
     size_t m_id;
+    Mod* m_mod;
+    bool m_inInterceptor;
```
dankmeme01: not initialized to `= false` so this will be undefined behavior on first read

loader/src/utils/web.cpp
```diff
@@ -355,15 +373,44 @@ class web::WebRequestsManager {
 
     struct RequestData {
         std::shared_ptr<WebRequest::Impl> request;
+        Mod* mod;
```
dankmeme01, line 358: is there any reason mod is stored separately, when it's already in the `request`?

dankmeme01 (changes requested):

loader/src/utils/web.cpp
SMJSProductions: WebRequest::Impl is only implemented after this so you can't access it here

altalk23: you could use a setter instead

loader/src/utils/web.cpp
```diff
@@ -734,40 +786,72 @@ std::atomic_size_t WebRequest::Impl::s_idCounter = 0;
 WebRequest::WebRequest() : m_impl(std::make_shared<Impl>()) {}
 WebRequest::~WebRequest() {}
 
-WebFuture WebRequest::send(std::string method, std::string url) {
+WebFuture WebRequest::send(std::string method, std::string url, Mod* mod) {
+    if (m_impl->m_inInterceptor) throw std::runtime_error("Cannot call send while inside an interceptor callback.");
```
altalk23: can we do smth else here instead of throwing? maybe result, maybe terminate

loader/include/Geode/utils/web.hpp
```diff
@@ -11,6 +11,7 @@
 #include <chrono>
 #include <optional>
 #include <string_view>
+#include <shared_mutex>
```
altalk23: any reason to include it in the header?

loader/src/utils/web.cpp
dankmeme01: result would be too hacky with webfuture but terminate would work

altalk23:
lgtm

PR #1698, opened 2026-02-06
Create new and improved CCScale9Sprite (geode::NineSlice) :3

Alphalaneous:
- Properly size sprite rects to prevent overlap (low opacity looks good!)
- Ability to set it to repeat instead of scale the center
- Ability to set a scale multiplier for the slices (no more contentSize*2 then scale/2 stuff needed)
- Doesn't recreate the sprite every time something changes (looking at you cocos)
- Supports negative content size (why not)
- Self contained. The batch node is manually visited. This allows for devs to add nodes to it and use layouts on it without issue. 
- Better Insets. Cocos uses a CCRect for insets, which required the you to use math :fearful: to use the right and top insets. This one doesn't require that. 
- People no longer have to accidentally depend on Happy Textures for good looking CCScale9Sprites :3

### DIFF
--- a/loader/src/ui/nodes/NineSlice.cpp
+++ b/loader/src/ui/nodes/NineSlice.cpp
@@ -75,6 +75,8 @@ void NineSlice::setup(Insets const& insets, CCRect const& rect) {
         m_impl->m_insets = {size.height/3, size.width/3, size.height/3, size.width/3};
     }
 
+    m_impl->m_batchNode->setID("slice-batch");
+
     setAnchorPoint({0.5f, 0.5f});
     setContentSize(m_impl->m_spriteRect.size);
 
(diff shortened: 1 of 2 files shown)


loader/src/ui/nodes/NineSlice.cpp
```diff
@@ -0,0 +1,383 @@
...
+    CCRect bottomCenterRect = {texRect.origin.x + texLeft, texRect.origin.y + texRect.size.height - texBottom, texCenterW, texBottom};
+    CCRect bottomRightRect = {texRect.origin.x + texRect.size.width - texRight, texRect.origin.y + texRect.size.height - texBottom, texRight, texBottom};
+
+    CCAffineTransform t = CCAffineTransformMakeIdentity();
+    if (!rotated) {
+        t = CCAffineTransformTranslate(t, 0, 0);
+    }
+    else {
+        t = CCAffineTransformTranslate(t, texRect.size.height, 0);
+        t = CCAffineTransformRotate(t, 1.57079633f);
```
dankmeme01: imo make this a bit more clear, like `M_PI / 2.f` or `std::numbers::pi / 2.f`

PR #1737, opened 2026-02-14
Update internal packages

hiimjasmine00:


loader/src/utils/base64.cpp
```diff
@@ -17,7 +17,7 @@ std::string base64::encode(std::span<std::uint8_t const> data, Base64Variant var
     std::string buffer;
     buffer.resize(simdutf::base64_length_from_binary(data.size(), opt));
     // returned size is the same as above
-    std::ignore = simdutf::binary_to_base64(data, buffer, opt);
+    std::ignore = simdutf::binary_to_base64(std::span(reinterpret_cast<const char*>(data.data()), data.size()), buffer, opt);
```
matcool, line 20: ?

hiimjasmine00: This is required on simdutf v8

PR #1750, opened 2026-02-15
Add controller sprites to keybind settings

hiimjasmine00:


loader/include/Geode/utils/Keyboard.hpp
```diff
@@ -113,6 +113,7 @@ namespace geode {
         
         static Result<Keybind> fromString(std::string_view str);
         std::string toString() const;
+        cocos2d::CCNode* getNode() const;
```
matcool: maybe this should be `createNode`, since its creating one every time its called, not just returning some stored member

loader/src/loader/SettingNodeV3.cpp
```diff
@@ -626,7 +626,12 @@ void KeybindSettingNodeV3::updateState(CCNode* invoker) {
     buttonMenu->removeAllChildren();
     for (size_t i = 0; i < m_currentValue.size(); i++) {
         auto& keybind = m_currentValue[i];
-        auto buttonSprite = createGeodeButton(keybind.toString().c_str(), true);
+        CCNode* buttonSprite;
+        if (keybind.key < 1000 || keybind.key > 2000) {
```
matcool, line 631: maybe add a comment, the magic numbers arent obvious at a glance

PR #1757, opened 2026-02-15
refactor: Refactor server JSON parsing to use `matjson::Serialize` sp…

skibiditymus27:
…ecializations and add `ServerDeveloper` struct.


qimiko:
this pull request makes too many unnecessary changes that makes it hard to review. ignoring that, i don't think there's any reason to change these classes to use the matjson serialization system at this time

PR #1763, opened 2026-02-15
add event for setting keybinds but with a timestamp

qimiko:
draft until we figure out a good solution for this (can't break abi ☹️)


hiimjasmine00:
my bad

PR #1765, opened 2026-02-16
Improve version fetching with fallback to GitHub API

staFF6773:
Updated version retrieval logic to check GitHub releases if Geode Index fails.

> [!NOTE]
> This change fixes the bug when installing geode on Linux.
> Version: Geode v5.0.0-beta.2

Screenshot
image


Cvolton:
the pr is merged now so the point is moot but im resending my message that i attempted to send like 4 days ago that apparently never made it through (except shorter because i had to retype the main points from memory)

i dont think this is necessary, all you'd have to do to make this work for this specific case is add `?prerelease=true` to the url but that may not always be desired behavior, as we generally dont want to install prereleases, so the better approach might be to just have it download the version from the specific release the script comes from, will wait for others to chime in

(others never chimed in because the message never got sent i hate github)

PR #1775, opened 2026-02-17
Add mod pinning support

TheSillyDoggo:
Add's a button to pin a mod, which shows it above every other mod (but still below geode)

TODO (for someone else :3):
add colours for non geode theme


HJfod:
LGTM

PR #1782, opened 2026-02-17
Keybind setting listener improvements

hiimjasmine00:
- Fix keybind setting listeners not stopping when they need to
- More robust keybind event filtering
- Added keybind priority


hiimjasmine00:
I hope I won't mark this for draft again

PR #1787, opened 2026-02-17
Update for IR processing in Bindings

9inefold:
Change related to [IR processing in bindings](https://github.com/geode-sdk/bindings/pull/1297). Doesn't change current builds.


matcool:
LGTM

PR #1799, opened 2026-02-20
make ModItem "Get" button become "View" after download

Zilko:
turip ip ip


hiimjasmine00:
ip ip ip tsha ik

tamalitotactico:
hola

PR #1809, opened 2026-02-21
Fix unfriendly keybind modifiers

Vinsterplays:
- store held down keybinds in unordered map
- release event will ignore modifiers if keybind was in map

Fixes:
- Releasing a modifier before the keybind key doesnt fire release event
- Setting the keybind key to only a modifier key will not fire release event
- Changing a keybind while holding the previous one down will not fire release event

### DIFF
--- a/loader/src/loader/LoaderImpl.cpp
+++ b/loader/src/loader/LoaderImpl.cpp
@@ -386,30 +386,29 @@ void Loader::Impl::populateModList(std::vector<ModMetadata>& modQueue) {
     }
 
     if (!m_keybindSettings.empty()) {
-        auto activeKeybinds = std::make_shared<std::unordered_map<KeybindSettingV3*, Keybind>>();
-
-        KeyboardInputEvent().listen([this, activeKeybinds](KeyboardInputData& data) {
+        KeyboardInputEvent().listen([this](KeyboardInputData& data) {
             Keybind keybind(data.key, data.modifiers);
             bool down = data.action != KeyboardInputData::Action::Release;
             bool repeat = data.action == KeyboardInputData::Action::Repeat;
-            
             if (down) {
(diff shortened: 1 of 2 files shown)


loader/src/loader/LoaderImpl.cpp
```diff
@@ -386,15 +386,30 @@ void Loader::Impl::populateModList(std::vector<ModMetadata>& modQueue) {
...
+        auto activeKeybinds = std::make_shared<std::unordered_map<KeybindSettingV3*, Keybind>>();
+
+        KeyboardInputEvent().listen([this, activeKeybinds](KeyboardInputData& data) {
             Keybind keybind(data.key, data.modifiers);
             bool down = data.action != KeyboardInputData::Action::Release;
             bool repeat = data.action == KeyboardInputData::Action::Repeat;
-            for (auto& setting : m_keybindSettings[keybind]) {
-                if (KeybindSettingPressedEventV3(setting->getModID(), setting->getKey()).send(keybind, down, repeat, data.timestamp)) {
-                    return ListenerResult::Stop;
+
```
qimiko: extra whitespace 💔

loader/src/loader/LoaderImpl.cpp
```diff
@@ -386,15 +386,30 @@ void Loader::Impl::populateModList(std::vector<ModMetadata>& modQueue) {
     }
 
     if (!m_keybindSettings.empty()) {
-        KeyboardInputEvent().listen([this](KeyboardInputData& data) {
+        auto activeKeybinds = std::make_shared<std::unordered_map<KeybindSettingV3*, Keybind>>();
```
qimiko: instead of making a shared pointer to an unordered map i think it'd be better to store the map (without the pointer) as a member in `LoaderImpl`

qimiko (changes requested):

PR #1810, opened 2026-02-21
Add geode::Button, a CCMenuItem(SpriteExtra) alternative

Alphalaneous:
CCMenuItems require CCMenus, which can be a little annoying at times. Why not have something that just works as a button?

- Create a button with a label, sprite, or node
- Can customize callbacks for select, unselect, and activate
- Can use custom animation actions for select and unselect
- Easily use existing animations, Scale and Move, from CCMenuItemSpriteExtra (Scale is default)
- State resetting, so no worries if your action modifies some attribute on the button
- Shared buttons
  - All buttons sharing a parent will automatically have their touch shared with each other
  - Works similar to how CCMenus worked, but handled on a per button basis

### DIFF
--- a/loader/include/Geode/ui/Button.hpp
+++ b/loader/include/Geode/ui/Button.hpp
@@ -1,0 +1,234 @@
+#pragma once
+
+#include <Geode/utils/ZStringView.hpp>
+#include <cocos2d.h>
+
+namespace geode {
+    class GEODE_DLL Button : public cocos2d::CCNodeRGBA, public cocos2d::CCTouchDelegate {
+    public:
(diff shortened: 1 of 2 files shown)


xblazegmd:
I like this. It's a bit sad for me to leave `CCMenuItemSpriteExtra` just like that lol (yes I am aware this is marked as an *alternative*, not a *replacement*), but it's a very good concept.

It's kinda like with `FLAlertLayer` when for simple enough things `FLAlertLayer` works just fine but for other stuff like custom popups `geode::Popup` is more simple to use/make. Oh well..

HJfod:
Mm interesting. Maybe one day we will replace every class in GD and Cocos2d :) I approve this

PR #1843, opened 2026-02-23
Add $HOME/.steam/debian-installation to linux install script

mrmelon54:
On my Debian-based (Linux Mint) system the native Steam client has two steamapps directories.

- `$HOME/.local/share/Steam/steamapps`
- `$HOME/.steam/debian-installation/steamapps`

A symlink exists from `$HOME/.steam/steam` to `$HOME/.steam/debian-installation` if that is preferred.

### DIFF
--- a/installer/linux/install.sh
+++ b/installer/linux/install.sh
@@ -71,7 +71,7 @@ find_gd_installation() {
     verbose_log "Searching for Geometry Dash..."
     local DATA_HOME="${XDG_DATA_HOME:-$HOME/.local/share}"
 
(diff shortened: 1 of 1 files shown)


Fleeym:
There's no Steam folder in $HOME/.local/share? The existing config should pick that up.

mrmelon54:
Geometry Dash is not in `$HOME/.local/share/Steam/steamapps/common` but the path does exist. I am not sure why this is?
It does appear in `$HOME/.steam/steam` which is symlinked to `$HOME/.steam/debian-installation`.
Adding either of these paths will resolve this issue.

Screenshot of the existing installer not finding Geometry Dash on my system.
image

PR #1844, opened 2026-02-23
need an id for this

BeatACVR:


BeatACVR:
am i a fucking moron

BeatACVR:
how did i mess up a PR with one line of code

altalk23:
apparently nothing in settings has an id, why should just this have it, ill ask lead chat

BeatACVR:
> apparently nothing in settings has an id, why should just this have it, ill ask lead chat

i would like it for a mod. i can add id's for everything else if you want

PR #1854, opened 2026-02-25
Add geode::SliderNode, a Slider alternative

Alphalaneous:
- Easier to use and more customizable than GDs Slider
- Min and Max controls
- Get and set by value and percent
- Built in Slider Bypass
- Read Only mode (for stuff like progress bars)
- Callbacks for sliding, clicking, and releasing the thumb
- Ability to link geode::TextInput and cocos2d::CCLabelBMFont to the slider to have them automatically update
- Slider snapping steps
- Content size dictates the groove size, which is a geode::NineSlice now

### DIFF
--- a/loader/src/ui/nodes/SliderNode.cpp
+++ b/loader/src/ui/nodes/SliderNode.cpp
@@ -1,9 +1,9 @@
-#include <Geode/ui/SliderControl.hpp>
+#include <Geode/ui/SliderNode.hpp>
 #include <Geode/Geode.hpp>
 
 using namespace geode::prelude;
 
-class SliderControl::Impl final {
+class SliderNode::Impl final {
 public:
     float m_min = 0.f;
     float m_max = 100.f;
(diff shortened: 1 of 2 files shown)


loader/include/Geode/ui/SliderControl.hpp
```diff
@@ -0,0 +1,176 @@
...
+         * Create a custom Slider
+         * 
+         * @param thumb The sprite file name or frame name for the thumb when not selected
+         * @param thumbSelected The sprite file name or frame name for the thumb when selected
+         * @param groove The sprite file name or frame name for the grove
+         * @param bar The sprite file name (cannot be a frame) for the progress bar
+         * @param callback The callback for when the thumb is dragged
+         * @param barOffset The size offset (will be inset for each side) for the bar to fit within the groove
+         */
+        static SliderControl* createCustom(ZStringView thumb, ZStringView thumbSelected, ZStringView groove, ZStringView bar, SliderCallback callback, cocos2d::CCSize const& barOffset = {2.f, 2.f});
```
Fleeym: Would it be better if this took CCSprite* params instead of ZStringViews? Atm it could be doing 1 pointless CCSprite::create.

loader/src/ui/nodes/SliderNode.cpp
Fleeym, line 1: This is probably a personal nitpick, but adding `this->` before member function calls is a bit easier to read.

loader/include/Geode/ui/SliderControl.hpp
Alphalaneous: Fair fair, gonna change this when I get home

loader/src/ui/nodes/SliderNode.cpp
Alphalaneous: Personally my nitpick goes the opposite direction and it makes code noisy for me haha

Fleeym:
You forgot the bar it's still a ZStringView

cgytrus:
i think SimpleSlider would be a better name because, even though it still sucks, at least its consistent with SimpleTextArea

regarding concerns with intellisense, i dont think itd matter as if you already know the class name youd know which one to start typing and pick, and if you dont, you wont easily figure it out without docs anyway because the options upon typing "Slider" would either be "Slider" and "SliderNode" or "Slider" with "SimpleSlider" possibly a bit lower

we should probably revisit ui node names in v6 or something anyway because a lot of them seem to be bad

Alphalaneous:
I completely disagree with the Simple naming scheme altogether. Especially since this provides more advanced functionality. So I think I will close this PR for now

PR #1860, opened 2026-02-26
Fix shutdown not triggering game exit event on macOS

hiimjasmine00:


dankmeme01:
why is ery verified again

PR #1878, opened 2026-02-28
Add the `Custom Web DNS Servers` setting

M336G:
This pull request adds  `Custom Web DNS Servers` setting below `Web DNS Server` in Geode's settings

image

_Changing DNS servers may only be really effective after a restart as it seems that resolved addresses are cached for the session_

### DIFF
--- a/loader/src/utils/web.cpp
+++ b/loader/src/utils/web.cpp
@@ -23,6 +23,7 @@
 #include <arc/sync/mpsc.hpp>
 #include <arc/net/TcpStream.hpp>
 #include <asp/collections/SmallVec.hpp>
+#include <asp/iter.hpp>
 #include <ca_bundle.h>
 #include <curl/curl.h>
 #include <sstream>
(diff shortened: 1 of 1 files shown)


MegaSa1nt:
</img>

dankmeme01:
only saw the PR now after refactoring all dns stuff oops :P

I think the setting is a great idea, but you'd need to change some code after my commit, inside `dnsProbe()` you can check if a custom server is set and then use  `qsox::IpAddress::parse` to convert into ip address structs, and put it in the global var.

M336G:
accidentally closed oops, gonna reopen once that's fixed

M336G:
> only saw the PR now after refactoring all dns stuff oops :P
> 
> I think the setting is a great idea, but you'd need to change some code after my commit, inside `dnsProbe()` you can check if a custom server is set and then use `qsox::IpAddress::parse` to convert into ip address structs, and put it in the global var.

alright i believe it should be fixed!

Cvolton:
one thing i wonder about is what happens when a kid puts a keysmash into the field

M336G:
> one thing i wonder about is what happens when a kid puts a keysmash into the field

It'll log a warning and default to what's set for `Web DNS Server` (as if `Custom Web DNS Servers` was empty)

loader/src/utils/web.cpp
```diff
@@ -1771,6 +1771,23 @@ arc::Future<> WebRequestsManager::Impl::dnsProbe() {
         m_wakeNotify.notifyOne();
     });
 
+    auto custom = Mod::get()->getSettingValue<std::string_view>("curl-custom-dns3");
+    if (!custom.empty()) {
+        auto adresses = qsox::IpAddress::parse(std::string(custom));
```
dankmeme01: IpAddress stores one address, so that won't work if you want multiple. Something like this should work to get a `std::vector<qsox::IpAddress>`:

```cpp
auto ips = asp::iter::split(custom, ',').filterMap([&](std::string_view s) {
    return qsox::IpAddress::parse(std::string{s}).ok();
}).collect();
```

(you may need to include `<asp/iter.hpp>`)

Then you can check if `ips` is empty (no valid address specified), and if it's not `std::move` it into the `DnsServer`

PR #1884, opened 2026-03-01
Make modify_cast more powerful

nytelytee:
The original `modify_cast` was only able to cast a non-modify class to a modify class.

This pull request enables it to do the reverse, cast from a modify class to a non-modify class as well.

It also enables it to cast between 2 modify classes.

### DIFF
--- a/loader/include/Geode/modify/Modify.hpp
+++ b/loader/include/Geode/modify/Modify.hpp
@@ -629,7 +629,7 @@ namespace geode::cast {
      * modify_cast<HookedGJBaseGameLayer*>(myHookedPlayLayer);
      */
     template<typename Target, typename Original>
-    constexpr Target custom_modify_cast(Original original) {
+    constexpr Target modify_cast(Original original) {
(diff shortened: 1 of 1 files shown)


nytelytee:
~~One additional question I forgot:
would you be fine with me also adding a `modify_typeinfo_cast` that will do a `typeinfo_cast` at the right place in the cast chain, so that one can also use this at runtime? I think that at each of the 3 branches, only 1 `typeinfo_cast` is necessary, when switching between base types. The requires from the static asserts in the branches can also be removed in that case, I think.~~

Also the void checks inside of the static asserts in the branches may be redundant given that the concept in the if constexpr already checks for that. ~~Hence, in the `typeinfo_cast` version, the entire static asserts could be removed altogether.~~

nytelytee:
Does Geode have a style guide for concept names? I named a concept `IsRealModify`, but am not sure if I should have used snake_case or not.

dankmeme01:
looking at source it's inconsistent but mostly PascalCase

PR #1890, opened 2026-03-03
Rework formatting for logs and fmt::format

Alphalaneous:
- Separate formatting from the log headers and source

- Runtime formatting for CCObjects
  - Provide tools for registering CCObjects to be formatted easily, even for your own mods
    - Generally useful for API mods that add UI elements and wanting to simply be able to format the object
    - This also brings the ability for CCObjects within containers to have their information shown in a lot as well
  - Uses a Wrapper class for CCObjects instead of just an (ugly) wrapCocosObj method to wrap them. Enabling using the fmt::formatter directly, and thus allowing us to parse for specifiers too. 
    - You can use format::wrap with a CCObject to format it outside of logs, such as `fmt::format("My Object: {}", format::wrap(CCString::create("Wow!")));`
- New formats for CCFloat, CCDouble, CCInteger, CCBool, CCString, CCSet, CCDictionary, CCTouch, CCFiniteTimeAction, and CCActionInterval
- Prettify container formats by default, use the `raw` specifier to disable prettifying: `log::info("{:raw}", this->m_myDict);`

**Note that I did not remove the `format_as` for CCObject, CCArray, and CCNode, even though they are now obsolete. As that would be considered a breaking change. These will still function the same if someone is calling them for any reason.**

### DIFF
--- a/loader/include/Geode/loader/Formatter.hpp
+++ b/loader/include/Geode/loader/Formatter.hpp
@@ -145,6 +145,9 @@ namespace geode::format {
     };
 
     class FormatBase {
+    public:
+        virtual ~FormatBase() = default;
+    private:
         virtual geode::Result<std::string> format(cocos2d::CCObject const* obj, std::string_view specifier) = 0;
         friend class FormatterImpl;
     };
@@ -162,7 +165,7 @@ namespace geode::format {
         geode::Result<std::string> format(cocos2d::CCObject const* obj, std::string_view specifier) override {
             if (!geode::cast::typeinfo_cast<T*>(obj)) return geode::Err("Incorrect Type");
             if (!m_formatter) return geode::Err("No formatter callback");
-            if (m_formatter) return geode::Ok(m_formatter(static_cast<T*>(const_cast<cocos2d::CCObject*>(obj)), specifier));
+            return geode::Ok(m_formatter(static_cast<T*>(const_cast<cocos2d::CCObject*>(obj)), specifier));
         }
 
         friend class FormatterImpl;
--- a/loader/src/loader/Formatter.cpp
+++ b/loader/src/loader/Formatter.cpp
@@ -1,4 +1,6 @@
+#include <Geode/utils/StringBuffer.hpp>
 #include <Geode/loader/Formatter.hpp>
+#include <ranges>
 
 using namespace geode::prelude;
 
@@ -12,11 +14,11 @@ public:
     }
 
     void addFormat(std::string_view name, std::unique_ptr<FormatBase> format) {
-        m_formatCallbacks.insert(m_formatCallbacks.begin(), std::move(format));
+        m_formatCallbacks.push_back(std::move(format));
     }
 
     Result<std::string> handleFormat(CCObject const* obj, std::string_view specifier) {
-        for (const auto& callback : m_formatCallbacks) {
+        for (const auto& callback : std::views::reverse(m_formatCallbacks)) {
             auto res = callback->format(obj, specifier);
             if (res) return Ok(res.unwrap());
         }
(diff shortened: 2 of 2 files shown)


loader/include/Geode/loader/Formatter.hpp
```diff
@@ -0,0 +1,252 @@
...
+
+        ~FormatType() = default;
+        FormatType(FormatCallback callback) : m_formatter(std::move(callback)) {}
+    private:
+        FormatCallback m_formatter;
+
+        geode::Result<std::string> format(cocos2d::CCObject const* obj, std::string_view specifier) override {
+            if (!geode::cast::typeinfo_cast<T*>(obj)) return geode::Err("Incorrect Type");
+            if (!m_formatter) return geode::Err("No formatter callback");
+            if (m_formatter) return geode::Ok(m_formatter(static_cast<T*>(const_cast<cocos2d::CCObject*>(obj)), specifier));
```
dankmeme01: the redundant `if (m_formatter)` here can be a bit confusing so i'd remove it

loader/include/Geode/loader/Formatter.hpp
```diff
@@ -0,0 +1,252 @@
...
+    }
+};
+
+namespace geode::format {
+    struct Wrapper {
+        Wrapper(cocos2d::CCObject* obj) : m_obj(obj) {}
+        cocos2d::CCObject* m_obj;
+    };
+
+    class FormatBase {
```
dankmeme01, line 147: since you use `unique_ptr`, I recommend adding a virtual destructor to this class. Most formatters probably won't have any state, but without a virtual destructor you make it so that their destructors never get to run at all which can cause subtle bugs

loader/src/loader/Formatter.cpp
```diff
@@ -0,0 +1,221 @@
...
+    format::registerFormat<CCBool>([] (CCBool* self) -> std::string {
+        return self->getValue() ? "true" : "false";
+    });
+
+    format::registerFormat<CCString>([] (CCString* self) -> std::string {
+        return fmt::format("\"{}\"", self->getCString());
+    });
+
+    format::registerFormat<CCArray>([] (CCArray* self, std::string_view specifier) -> std::string {
+        if (!self && !self->count()) return "[empty]";
```
dankmeme01: this check sounds erroneous, did you mean `!self || !self->count()`

loader/src/loader/Formatter.cpp
```diff
@@ -0,0 +1,221 @@
...
+            buffer.append(fmt::format("{}", format::wrap(obj)));
+        }
+
+        if (!rawFormat) buffer.push_back('\n');
+        buffer.push_back(']');
+        return fmt::to_string(buffer);
+    });
+
+    format::registerFormat<CCSet>([] (CCSet* self, std::string_view specifier) -> std::string {
+        if (!self && !self->count()) return "[empty]";
```
dankmeme01: ditto

loader/src/loader/Formatter.cpp
```diff
@@ -0,0 +1,221 @@
...
+private:
+    std::vector<std::unique_ptr<FormatBase>> m_formatCallbacks;
+public:
+    static FormatterImpl* get() {
+        static auto s_instance = new FormatterImpl();
+        return s_instance;
+    }
+
+    void addFormat(std::string_view name, std::unique_ptr<FormatBase> format) {
+        m_formatCallbacks.insert(m_formatCallbacks.begin(), std::move(format));
```
dankmeme01: is there a specific reason why this inserts at the beginning (which is going to be O(n) always)? if this is to make new listeners highest priority, IMO just iterate in reverse when calling

loader/src/loader/Formatter.cpp
```diff
@@ -0,0 +1,221 @@
...
+            if (!first) buffer.append(std::string_view(", "));
+            if (!rawFormat) buffer.append(std::string_view("\n\t"));
+            first = false;
+
+            buffer.append(fmt::format("{}", format::wrap(obj)));
+        }
+
+        if (!rawFormat) buffer.push_back('\n');
+        buffer.push_back(']');
+        return fmt::to_string(buffer);
```
dankmeme01: I would recommend using `geode::utils::StringBuffer` (in here and all other functions), it's going to be more efficient and less verbose to use (e.g. you can directly do `buffer.append(", ")` or `buffer.append("{}", format::wrap(obj))`

PR #1893, opened 2026-03-04
Fix small issue in UI overlap between 'Outdated' and 'Restart Required' labels

ANGELUSD11:
Adjusted the Y offset for the m_outdatedLabel in all three display modes (Grid, SmallList, and BigList) to stack the labels vertically instead of anchoring them to the exact same coordinates. This is a minimal and non-invasive fix to prevent the text collision.


Cvolton:
thanks for the pr, unfortunately we ended up opting for a different approach to solving the issue in 70e0fd6034a2899554cb6d6077407a6874efbd3a

PR #1921, opened 2026-03-15
Add malware warning

TheSillyDoggo:
"Geode could not find the main function, not loading Geode.
Your PC is very likely infected with malware. It's recommended you reinstall Windows."

### DIFF
--- a/loader/src/platform/windows/main.cpp
+++ b/loader/src/platform/windows/main.cpp
@@ -291,7 +291,7 @@ std::string loadGeode() {
     }
 
     if (patchAddr == 0)
-        return "Geode could not find the main function, not loading Geode.";
+        return "Geode could not find the main function, not loading Geode.\nYour PC is very likely infected with malware. It's recommended you reinstall Windows.";
 
 #define JMP_ADDR(from, to) (std::bit_cast<uintptr_t>(to) - std::bit_cast<uintptr_t>(from) - 5)
 #define JMP_BYTES(from, to) \


qimiko:
i don't like the fear this warning would create, would definitely reword it to be closer to something like "your gd installation is likely corrupt/invalid/illegitimate/idk"

Alphalaneous:
> i don't like the fear this warning would create, would definitely reword it to be closer to something like "your gd installation is likely corrupt/invalid/illegitimate/idk"

Disagree, because 99/100 times it is gonna be due to malware. Better to at least make the end user aware.

Cvolton:
also not sure about the directness of the message but i personally def agree with the spirit of the pr

Cvolton:
one more thing that we'd need to make sure is that if a message like this were to be added, it can't make people think that geode is how they got infected

xblazegmd:
Maybe *"It is recommended you check for malware on your device"* or smth along those lines, idrk

Iliya-code:
I have an idea. "Geode could not find the main function, not loading Geode. Scan your system for malware or reinstall Geode."

FireMario211:
> I have an idea. "Geode could not find the main function, not loading Geode. Scan your system for malware or reinstall Geode."

Reinstalling Geode would do nothing because the issue only happens when it can't find the main function **in Geometry Dash** 😭

xblazegmd:
Ooh I have an idea:
*"Failed to find the main function, not loading Geode. Please reinstall Geometry Dash and/or check for malware on your device"*

xblazegmd:
Or maybe *"run an antivirus on your device"* if we wanna be more soft

Alphalaneous:
The malware that causes this won't be able to be removed that easily. Usually requires a full reinstall of Windows since it infects everything

FireMario211:
> Ooh I have an idea: _"Failed to find the main function, not loading Geode. Please reinstall Geometry Dash and/or check for malware on your device"_

@xblazegmd Like alpha said, the malware likely has infected other executable files as well, not just GD. An antivirus wouldn't solve it because it would mark every single program as infected, which is why the initial message in the PR said to reinstall windows.

Though I have to agree with Cvolton because people will think their Geode was infected, so perhaps rewording it to mention that the malware they have has infected their other programs as well, and happened before they ran Geode. Something like that.

xblazegmd:
Maybe if it's too hard to word out in a single sentence there could be a separate "more info" page for it?

So it would look like
*"Failed to find the main function, not loading Geode. Please scan for malware on your device. Learn more: geode-sdk.org/some/page/abt/malware"*

It also would be useful to clear up some stuff to the user like the fact Geode is not how they got infected, how to be 100% sure it is malware, etc

xblazegmd:
Only concern is the most likely tons of 5 year olds that would get too scared and not go to the page

xblazegmd:
So I need to think more abt how to word it out

xblazegmd:
Ok maybe:
*"Geode could not find the main function, not loading Geode.
This was most likely caused by malware on your device. It is recommended you reinstall Windows. Learn more: geode-sdk.org/some/info/page)"*

Then the page would, as I said, go more in detail abt the malware issue and say smth along the lines of:
*"This means your computer has most likely been infected with malware. This form of malware infects other executable files with its own malicious code, that then spreads to even more executable files, until your whole system is infected with the malware. This issue most likely predates the installation of Geode. It is recommended you reinstall Windows"*

I still have the same concern of ppl not going to the page (maybe it could be all in another popup?), however there is only so much you can word out on a single error message

PR #1931, opened 2026-03-16
add `geode::Dropdown` and replace `one-of` arrows with dropdown

TheBearodactyl:


Cvolton:
why does this pr contain so many unrelated formatting changes

TheBearodactyl:
because of alk
image

dankmeme01:
don't think anybody ever runs a formatter in this repo 💀

TheBearodactyl:
me, i do
i have format on save enabled
and sadly i can't exactly just unformat it

HJfod:
I don't have a strong opinion on including this util in Geode but I strongly disagree with using it anywhere in the loader UI since it is not a UX pattern found elsewhere in vanilla GD

Alphalaneous:
> I don't have a strong opinion on including this util in Geode but I strongly disagree with using it anywhere in the loader UI since it is not a UX pattern found elsewhere in vanilla GD

Disagree to an extent. It's easier to know what options are available for one-of. But I also feel it shouldn't be a straight up replacement for the arrows, and instead a control setting

HJfod:
> > I don't have a strong opinion on including this util in Geode but I strongly disagree with using it anywhere in the loader UI since it is not a UX pattern found elsewhere in vanilla GD
> 
> Disagree to an extent. It's easier to know what options are available for one-of. But I also feel it shouldn't be a straight up replacement for the arrows, and instead a control setting 

I agree from a general UX standpoint this would pretty obviously be better, but I don't want Geode UI to stray away from GD's established UX. Geode already doesn't look like GD, so it should at least feel like it. If we add whatever, then the cognitive load we dump on our potentially non-English-speaking non-computer-savvy 10-year-old children will be too much imo.

Of course this is a very simple component and one they should surely be familiar with from elsewhere, but I want to keep it as a hard rule that Geode may not introduce new UX interactions unless absolutely necessary and if the provided UX adds a considerable benefit.

Alphalaneous:
Personally I feel like that being a hard rule really limits potential. It is a very simple component as you say. There's a certain level of babying the users that just feels unreasonable here.

Fleeym:
I personally think one-of settings are extremely annoying to use at the moment, since you have no way of knowing how many values there are until you scroll the entire list through. Having a dropdown would make it so much nicer to use.

Dropdowns are such a widely used pattern on the web that I doubt someone has never heard of one or how it works. In my opinion sticking as close as possible to vanilla should be a balancing act, not a hard rule. If the benefits outweigh the cons by a lot, why not add it?

As long as it's large enough to be easily used on mobile I'd say it's an improvement.

Some screenshots with the dropdown in-game would be useful here.

HJfod:
I agree one-of settings are currently pretty ass, but I'd rather add a popup with a list of options and a "Use" button (similar to song selection and other popups in the editor) rather than dropdowns

Fleeym:
> I agree one-of settings are currently pretty ass, but I'd rather add a popup with a list of options and a "Use" button (similar to song selection and other popups in the editor) rather than dropdowns

That could work, though it's out of the scope of this PR.

TheBearodactyl:
> > I don't have a strong opinion on including this util in Geode but I strongly disagree with using it anywhere in the loader UI since it is not a UX pattern found elsewhere in vanilla GD
> 
> Disagree to an extent. It's easier to know what options are available for one-of. But I also feel it shouldn't be a straight up replacement for the arrows, and instead a control setting

will make it into an optional setting, like how you can enable slider control on number settings

FireMario211:
Will repeat what I said earlier but using this class, the scrolling (specifically touch) doesn't work at times, other than mouse scrolling that is.

TheBearodactyl:
and like i said on discord, i'll fix that later, like sometime in the next 3 days

PR #1940, opened 2026-03-20
add restart in safe mode btn to windows crash popup

Cvolton:
kind of a primitive change but I opted to make this a PR instead of pushing this directly primarily because of the added `geode::utils::game::restart` overload which we would have to carry for the rest of v5 if we suddenly decide we don't like it lol

### DIFF
--- a/loader/src/platform/windows/util.cpp
+++ b/loader/src/platform/windows/util.cpp
@@ -282,7 +282,10 @@ void geode::utils::game::restart(bool saveData, bool safeMode) {
 
     wchar_t buffer[MAX_PATH];
     GetModuleFileNameW(nullptr, buffer, MAX_PATH);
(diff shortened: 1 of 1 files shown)


Cvolton:
Please see the following attachment for a preview of the updated popup.

image

loader/src/platform/windows/util.cpp
```diff
@@ -278,7 +282,7 @@ void geode::utils::game::restart(bool saveData) {
 
     wchar_t buffer[MAX_PATH];
     GetModuleFileNameW(nullptr, buffer, MAX_PATH);
-    auto const gdName = L"\"" + std::filesystem::path(buffer).filename().native() + L"\"";
+    auto const gdName = L"\"" + std::filesystem::path(buffer).filename().native() + L"\" --geode:safe-mode";
```
matcool: forgot to check if safeMode is actually true

Cvolton: :(

PR #1949, opened 2026-03-23
set dll overrides inside the installer using registry without needing launch options

quadruplea0:
yes and also made the ascii thing look better cause it was pissing me off


Fleeym:
Not sure if I like this approach since it makes it harder to remove the override.

This also makes wine required on the users path, which not all users will have.

Lastly, I might be wrong here but deleting the prefix will remove the override.

quadruplea0:
> Not sure if I like this approach since it makes it harder to remove the override.
> 
thats true, safe mode also exists tho but its kinda unpractical. you can just move the xinput dll tho if you really need to.
> This also makes wine required on the users path, which not all users will have.
> 
i can try to use proton instead but it might be kinda finnicky
> Lastly, I might be wrong here but deleting the prefix will remove the override.
>
i mean it uses gd's prefix, so the override will only be removed when you uninstall gd most of the time, i dont think anyone is going out of their way to delete gds wineprefix.

PR #1952, opened 2026-03-24
(less scary) don't save mod jsons if they are empty

Cvolton:
This PR is scary.

The motivation is primarily Android's power management - with about 180 mods installed saving takes about 3 seconds in the current state, which means there is less time to save the actual GD save file (which the user usually cares about more) before phones kill the app. Moving mod saves after GD introduces the opposite issue though - GD takes ages to save and Geode settings often don't save at all. This PR aims to reduce the time spent on saving mod data which hopefully increases the chance to successfully save normal GD data.

This is a functional implementation primarily made as a proof of concept. My game currently takes 0.28s to save mod data with 182 mods installed on android64 with this implementation.

### DIFF
--- a/loader/include/Geode/loader/ModSettingsManager.hpp
+++ b/loader/include/Geode/loader/ModSettingsManager.hpp
@@ -67,7 +67,5 @@ namespace geode {
          * for this mod, they are also reloaded for the dependant mods
          */
         void addDependant(Mod* mod);
-
-        bool isDefaultValues() const;
     };
 }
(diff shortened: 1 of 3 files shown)


Cvolton:
Reverted this to an older commit, since the default mod.json save skip changes how changing defaults in mod updates is handled and introduces inconsistent behavior there. The current version is slower than the most optimized one but it still introduces a significant performance gain (1 vs 3 seconds in my testing)

PR #1964, opened 2026-03-29
NodeEvent, MenuItemActivatedEvent

Alphalaneous:
Uses the Cocos Script Engine system to add new events for CCNodes and CCMenuItems. Since Geometry Dash never sets these up itself, we can make our own and use them for useful events. None of these require hooks.

### NodeEvent
- Can be filtered by `OnEnter`, `OnExit`, `OnEnterTransitionDidFinish`, `OnExitTransitionDidStart`, and `OnCleanup`, which match their corresponding methods.
- Has helper methods in CCNode.h to add callbacks easily
- Is global in case a developer wants to listen to every time these are called (Me, I am that developer)

### MenuItemActivatedEvent
- Will run when a CCMenuItem is pressed (after the callback)
- Has a helper method in CCMenuItem.h so you can add callbacks easily to menu items.

**If this PR is merged and a Geode Update is prepared, it will need to be synchronized with an Alpha's Geode Utils update, or else things will break.**


dankmeme01 (commented):
this is hacky and i love it!

most of the PR looks fine to me, but there is one concern i have with `ObjectDestroyedEvent`. does this end up running on every CCObject *ever* created (and destroyed)? because there are multiple potential issues with this:

1. thread safety - it is common for things like CCImage, CCTexture2D, etc. to be created in a thread. if this event fires for those, that would likely cause UB in the current implementation, as you are not using `ThreadSafeGlobalEvent`
2. performance hit - going through the event machinery on every destroyed `CCObject` sounds like it can have real performance implications, and this will especially be worse if thread-safe events are used. would need to benchmark it by creating & deleting a huge amount of objects and comparing the time taken with & without the event being sent

dankmeme01 (approved):

loader/src/loader/ScriptEngine.cpp
```diff
@@ -0,0 +1,45 @@
...
+        int executeLayerKeypadEvent(CCLayer* layer, int eventType) { return -1; }
+        int executeAccelerometerEvent(CCLayer* layer, CCAcceleration* accelerationValue) { return -1; }
+        int executeEvent(int handler, const char* eventName, CCObject* eventSource, const char* eventSourceClassName) { return -1; }
+        int executeEventWithArgs(int handler, CCArray* args) { return -1; }
+        bool handleAssert(const char* msg) { return false; }
+        bool parseConfig(ConfigType type, const gd::string& str) { return false; }
+    };
+}
+
+$execute {
```
altalk23: use on mod loaded

loader/src/loader/ScriptEngine.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+namespace geode {
+    class ScriptEngine : public CCScriptEngineProtocol {
+
+        ccScriptType getScriptType() {
+            // Javascript engine is called without needing to change m_nLuaID
+            return cocos2d::ccScriptType::kScriptTypeJavascript;
+        }
+        
+        int executeNodeEvent(CCNode* node, int action) {
```
altalk23, line 13: have you benchmarked

altalk23 (approved):

altalk23:
wrong button oops

loader/src/loader/ScriptEngine.cpp
altalk23: actually its fine if dank approved

PR #1971, opened 2026-03-31
Add browser link validation to openLinkInBrowser

qimiko:
renames the per platform `openLinkInBrowser` to `openLink` and then makes `openLinkInBrowser` a wrapper to `openLink` that validates if the passed in url is a link that would open in browser

parsing is bad but that's okay

### DIFF
--- a/loader/src/utils/web.cpp
+++ b/loader/src/utils/web.cpp
@@ -1833,42 +1833,22 @@ arc::Future<> WebRequestsManager::Impl::dnsProbe() {
 
 void utils::web::openLinkInBrowser(ZStringView url) {
     // evil path check for windows
-    if (url.view().contains("#:")) {
+    auto urlView = url.view();
+    if (urlView.contains("#:")) {
         return;
     }
 
-    // libcurl doesn't handle mailto, but mailto links are a simpler case
-    if (utils::string::toLower(url).starts_with("mailto:")) {
-        openLinkUnsafe(url);
+    auto schemeEnd = urlView.find_first_of(':');
+    if (schemeEnd == std::string_view::npos) {
         return;
     }
 
-    auto h = curl_url();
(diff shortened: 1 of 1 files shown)


loader/include/Geode/utils/web.hpp
```diff
@@ -16,6 +16,17 @@
 #include <span>
 
 namespace geode::utils::web {
+    /**
+     * Opens an arbitrary link, which does not have to be a web URL.
+     *
+     * Warning: do not use this on untrusted user input! This can lead to remote code execution.
+     */
+    GEODE_DLL void openLink(ZStringView url);
```
matcool: openLinkUnsafe would be a better name maybe? ppl will accidentally use this without reading the comments i feel..

loader/src/utils/web.cpp
```diff
@@ -1829,4 +1829,46 @@ arc::Future<> WebRequestsManager::Impl::dnsProbe() {
...
+}
+
+void utils::web::openLinkInBrowser(ZStringView url) {
+    // evil path check for windows
+    if (url.view().contains("#:")) {
+        return;
+    }
+
+    // libcurl doesn't handle mailto, but mailto links are a simpler case
+    if (url.view().starts_with("mailto:")) {
```
Alphalaneous: These are all case insensitive, you should be making the url lowercase beforehand

loader/include/Geode/utils/web.hpp
qimiko: good idea i renamed it

loader/src/utils/web.cpp
qimiko: `curl_url_get` returns a lowercase scheme even if an uppercase one is provided, but i did add the lowercase check for mailto

PR #1986, opened 2026-04-01
Fix German translation for installer

hhappa0-code:
## Summary
This PR refines the German translation strings for the installer and uninstaller to sound more natural.

## Changes
- **Terminology:** Changed "nicht rennt" (incorrect literal translation) to "geschlossen".
- **Translation:** Translated 2 strings that haven't been translated.
- **Clarity:** Changed "Pfad" to "Ordner". (sounds more natural)
- **Formatting:** Added proper German hyphenation for "Geometry Dash-Version".

### DIFF
--- a/installer/windows/Language Files/GermanExtra.nsh
+++ b/installer/windows/Language Files/GermanExtra.nsh
@@ -1,18 +1,18 @@
 !insertmacro LANGFILE_EXT German
 
 !pragma warning disable 6030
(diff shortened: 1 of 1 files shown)


Cvolton:
I'm merging anyway since this is an easy fix to do myself but the ` (the dll trademark)` text should be kept in the string, it's replaced by the installer with the name of the actual conflicting mod dll found

hhappa0-code:
Can believe I've missed that. Thank you!

PR #1993, opened 2026-04-02
Mod::getHash(), mod hashes in crashlog and removal of some leftover sapphire references

Uproxide:
those who hash. get hashed. brown!
image


Uproxide:
1993 was a good year

PR #1999, opened 2026-04-03
Add padding to AxisLayout and SimpleAxisLayout

prevter:
Adds optional padding settings into layouts, to offset the children inside it

image

(i will push DevTools changes when this gets merged for obvious reasons)


Cvolton:
it has the green checkmark so it's good imo

PR #2010, opened 2026-04-06
add "Hide Name in Crashlogs" option

TheSillyDoggo:
Tested on macOS and Linux (proton)
image

image


matcool:
replace with `<user>` or even `***` imo.. `{USERNAME}` is noisy

TheSillyDoggo:
changed to "\<user\>"

loader/src/internal/crashlog.cpp
```diff
@@ -268,6 +268,23 @@ std::string crashlog::writeCrashlog(
     // this could also be done by saving a loader setting or smth but eh.
     (void)utils::file::writeBinary(crashlog::getCrashLogDirectory() / "last-crashed", {});
 
+    auto newInfo = std::string(info);
+    auto newStack = std::string(stacktrace);
+
+    #if defined(GEODE_IS_WINDOWS) || defined(GEODE_IS_MACOS)
+    if (Mod::get()->getSettingValue<bool>("hide-user-in-crashlogs")) {
+        const char* user = std::getenv("USER");
```
qimiko, line 276: i don't think `USER` is defined on windows, you might be better off using the apis provided by the os like `GetUserName` 
(if you had objective c you could use `NSUserName` on mac but unfortunately you are stuck with environment variables)

altalk23: mac has objc crashlog
actually why not add it as geode util
getLocalUsername

TheSillyDoggo: i did env variable so i could hide linux username, the winapi function just returns "steamuser"

altalk23: if its not useful to expose it can be a private func

altalk23: rather internal util

qimiko: we do have a way to check for wine, so if you want that behavior (which makes sense to do on linux) then i'd only do it when wine is detected
also as alk said, this might be best as a proper geode util, which we definitely would want working correctly on both wine and non-wine gd

dankmeme01:
the entire way of replacing the username is a bit hacky, afaik there's only two places where the username can appear so it's cleaner to just sanitize these parts instead of running find and replace on the entire text.

I would suggest modifying `CrashContext::initialize` to just iterate over all images in the end and remove username from the path, and then do the same for stack frames and their `file` member

loader/src/internal/crashlog.cpp
Cvolton: `USER` env variable is not defined on native Windows, `USERNAME` is (which should return the same thing as GetUserName, which means `steamuser` in Wine)

The current approach won't work on native Windows. Not personally sure if we should have a Geode util return the Linux username inside of Wine, I feel like getting steamuser would be more expected, though I can see that returning the Linux username would be more useful for this use case.

PR #2019, opened 2026-04-09
Separate geode::Priority into a new header

hiimjasmine00:
Because it has more uses than hook priority now, I guess.


ArianaGrandeNatural:
New check passed

PR #2033, opened 2026-04-16
Add launch method for iOS (For crash-logs and other)

FireMario211:
not sure if the names should be changed or have different if conditions


Cvolton:
its probably good but i know nothing about ios thanks for reading my pr review

FireMario211:
thank you very much cvolton amoungus

PR #2040, opened 2026-04-18
Fixed touch delegate leak in RichTextArea

abb2k:
so recently i downloaded Better Touch Prio and it started alerting me of leaks, and so i checked and realized i should probably register touch in onEnter and unregister on onExit 😬


abb2k:
ya check failed for download error so i re-ran it :) now its good !!

PR #2082, opened 2026-05-23
Fix windows startup freeze when geode fails to load

pipe1os:
This pr fixes a startup freeze when geode fails to load early.

if an early error happens, earlyError was calling console::messageBox directly 
inside dllmain. problem is dllmain runs under the windows loader lock so 
calling ui there just freezes the process entirely instead of showing 
the error.

my fix is pretty simple, just store the message and show it later in 
gdMainHook once the loader lock is released. if hooking fails completely 
and gdMainHook never runs, the error still gets written to 
_geode_early_error.txt so it won't be a silent hang either way.


dankmeme01:
are there examples of circumstances where this actually happens? as far as I am aware, the loader lock is mostly a legacy thing, and nowadays you can do a lot of things in DllMain even if microsoft discourages it.

unless this is a real issue that happens with some setups, this doesn't sound like a very useful fix

pipe1os:
> are there examples of circumstances where this actually happens? as far as I am aware, the loader lock is mostly a legacy thing, and nowadays you can do a lot of things in DllMain even if microsoft discourages it.
> 
> unless this is a real issue that happens with some setups, this doesn't sound like a very useful fix

i don't have a repro case for this, but in my understanding, loader lock isn't a legacy thing. microsoft still strictly forbids calling user32/ui functions like messagebox inside dllmain because it can cause deadlocks if user32.dll isn't already loaded when dllmain runs.

if that happens during an early error, instead of showing the error dialog the process just freezes silently. if you think this is too much of an edge case to merge, that's ok.

dankmeme01:
hmm alright, it's just weird to fix this when I don't think anyone has ever had a problem with it. but it probably isn't a bad idea.

though when it comes to the implementation, it looks like you return and completely prevent the game from loading if an early error occurs, which is different from what we did before (simply not loading Geode). this probably should be fixed, and it's a good idea to fake some error to check it works as intended.

pipe1os:
I realized that earlyError only gets called when the hook fails, so gdMainHook never runs and s_earlyErrorMsg never gets read.

so i just do the same thing upgradeThread already does and run the messagebox in a thread. that way it runs after dllmain returns and the loader lock is gone.

the return FALSE case is separate since the dll is about to unload anyway, thread would be unsafe there so just call it directly.

tested by throwing a fake earlyError at the top of DLL_PROCESS_ATTACH, dialog showed up fine and gd ran vanilla after closing it.

PR #2103, opened 2026-06-22
This was bothering me too much

SMJSProductions:


xblazegmd:
Such a revolutionary change, Geode will never be the same after this

SMJSProductions:
I know
Geode will no longer trigger my OCD!
image

Stazzical:
I would've done the same

PR #2104, opened 2026-06-23
Bug Fixes

SMJSProductions:
This is so that custom UIs don't have issues doing things like installing required mods in mass due to the lack of access to the UIs which provide this within Geode.


SMJSProductions:
Repurposed the PR for bug fixes which I fixed while making the download tools I later scrapped.

PR #2112, opened 2026-07-03
Ambigious ctor

SMJSProductions:
The main copy CTOR already allows CCPoint being used to construct CCSize, making it ambiguous to use when providing an initializer list (e.g. m_contentSize({ 0, 0 }))


SMJSProductions:
Apparently I forgot to merge master so ignore the random extra commit

loader/src/ui/nodes/Scrollbar.cpp
Cvolton, line 204: looks good

PR #2117, opened 2026-07-08
Added color codes

SMJSProductions:


loader/include/Geode/utils/ColorCode.hpp
```diff
@@ -0,0 +1,92 @@
+#pragma once
+
+#include <ccTypes.h>
```
dankmeme01, line 4: missing includes, some things might be implicitly included here but `<string_view>` and `<Geode/Result.hpp>` are for sure missing

loader/include/Geode/utils/ColorCode.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+namespace geode {
+    template<typename T>
+    concept has_4b = requires(T object) { object.setColor(cocos2d::ccColor4B()); };
+    template<typename T>
+    concept has_3b = requires(T object) {
+        object.setColor(cocos2d::ccColor3B());
+        object.setOpacity(GLubyte());
+    };
```
dankmeme01, line 12: weird naming for something that's globally available in the geode namespace, i would go with `CanSetColor4B` and `CanSetColor3B`

loader/include/Geode/utils/ColorCode.hpp
```diff
@@ -0,0 +1,92 @@
...
+        /// Parses the given hex string using one of the following formats:
+        /// #R/G/B, #RR/GG/BB, #RGB, #RGBA, #RRGGBB or #RRGGBBAA
+        /// Where the # prefix is optional and / means it's 1 byte/nibble representing all channels
+        /// @returns The parsed result
+        static Result<Color> parse(std::string_view hex);
+    private:
+        inline static GLubyte getByte(uint32_t colorNum, size_t index, bool isShort);
+    public:
+        /// @returns The current alpha value
+        operator float() const;
```
dankmeme01: i'm 50/50 on this one - if someone does `float x = color;` there is virtually no way for them to see this comment. it's not immediately obvious what exactly this getter will return, and it might lead to accidental mistakes, when someone passes a color to a float function without even realizing it.

so i would personally remove

loader/include/Geode/utils/ColorCode.hpp
```diff
@@ -0,0 +1,92 @@
...
+        operator cocos2d::ccColor3B() const;
+        operator cocos2d::ccColor4B() const;
+        operator cocos2d::ccColor4F() const;
+        operator cocos2d::extension::HSV() const;
+        Color& operator=(Color&& other) = default;
```
dankmeme01: move assignment operators and constructors shall be noexcept (copy can also be in this case, but not as important)

loader/include/Geode/utils/ColorCode.hpp
```diff
@@ -0,0 +1,92 @@
...
+        Color& operator=(const Color& other) = default;
+
+        GLubyte r;
+        GLubyte g;
+        GLubyte b;
+        GLubyte a;
+
+        Color(Color&& other) = default;
```
dankmeme01: same as above

loader/include/Geode/utils/ColorCode.hpp
```diff
@@ -0,0 +1,92 @@
...
+        void setGreen(GLubyte green);
+        GLubyte getGreen() const;
+        void setB(GLubyte blue);
+        GLubyte getB() const;
+        void setBlue(GLubyte blue);
+        GLubyte getBlue() const;
+        void setA(GLubyte alpha);
+        GLubyte getA() const;
+        void setAlpha(GLubyte alpha);
+        GLubyte getAlpha() const;
```
dankmeme01: i personally find those methods to be useless as all fields are public, so it's easier and clearer to do `color.a` than `color.getA();`. there couldn't be any side effects to getting/setting them anyway

loader/include/Geode/utils/ColorCode.hpp
```diff
@@ -0,0 +1,92 @@
...
+        cocos2d::ccColor3B to3B() const;
+        cocos2d::ccColor4B to4B() const;
+        cocos2d::ccColor4F to4F() const;
+        cocos2d::extension::HSV toHSV() const;
+        /// @returns True if alpha is 0
+        bool isInvisible() const;
+        /// @returns True if alpha is bigger than 0 and smaller than 255
+        bool isTransparent() const;
```
dankmeme01: where does this definition of transparency come from, why are all values besides 0 and 255 considered transparent?

i'm not sure if this method is necessary to begin with, but the way i'd name these 3 functions is:

* isInvisible -> isTransparent
* isTransparent -> hasAlpha? or remove entirely, not sure what purpose this serves
* isOpaque -> isOpaque

loader/src/utils/ColorCode.cpp
```diff
@@ -0,0 +1,277 @@
+#include <Geode/utils/ColorCode.hpp>
+
+using namespace geode::prelude;
+
+Result<Color> Color::parse(std::string_view hex) {
+    const bool isPrefixed = hex.starts_with("#");
+
+    if (hex.size() == isPrefixed) {
+        return Err("Empty color string");
+    }
```
dankmeme01: i see what this condition does, but it's cryptic and hard to understand. it's enough to just do:

```cpp
if (hex.starts_with("#")) {
    hex.remove_prefix(1);
}
```

this way `hex` now contains a version without the hash

loader/src/utils/ColorCode.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+    const std::string_view hexStr = hex.substr(isPrefixed);
+    const size_t size = hexStr.size();
+
+    // If the string is not hex and the size is not an exact size of 1, 2, 3, 4, 6 or 8
+    if (hexStr.find_first_not_of("0123456789ABCDEFabcdef") != std::string_view::npos) {
+        return Err("Non base16 characters");
```
dankmeme01: this check is unnecessary since numFromString already will return an error if there are invalid chars

loader/src/utils/ColorCode.cpp
```diff
@@ -0,0 +1,277 @@
...
+    } else if (size > 4 && size != 6 && size != 8) {
+        return Err("Unsupported size");
+    }
+
+    GEODE_UNWRAP_INTO(const uint32_t colorNum, utils::numFromString<uint32_t>(hexStr, 16));
```
dankmeme01: i'm not entirely sure, but it's worth testing that this really *does* work for capital letters. we use [`fast_float`](https://github.com/fastfloat/fast_float) and i would assume it treats the `base` argument as a generic value so it doesn't magically assume to treat uppercase chars same as lowercase ones.

if that is the case, it's better to replace `numFromString` with a simple local hexDecode function

loader/src/utils/ColorCode.cpp
```diff
@@ -0,0 +1,277 @@
...
+        color.g = Color::getByte(colorNum, 2, isShort);
+        color.b = Color::getByte(colorNum, 1, isShort);
+        color.a = Color::getByte(colorNum, 0, isShort);
+    }
+
+    return Ok(Color(color));
+}
+
+GLubyte Color::getByte(uint32_t colorNum, size_t index, bool isShort) {
+    const GLubyte byte = colorNum >> (isShort ? 4 : 8) * index & 0xFF;
```
dankmeme01: this might work fine but i would add more parentheses so the operator precedence is clear.

or an even better solution is to do more local variables instead of a big expression

loader/src/utils/ColorCode.cpp
```diff
@@ -0,0 +1,277 @@
...
+Color::operator ccColor3B() const {
+    return { r, g, b };
+}
+
+Color::operator ccColor4B() const {
+    return { r, g, b, a };
+}
+
+Color::operator ccColor4F() const {
+    return { static_cast<GLfloat>(r / 255.0), static_cast<GLfloat>(g / 255.0), static_cast<GLfloat>(b / 255.0), static_cast<GLfloat>(a / 255.0) };
```
dankmeme01: i think this can be greatly simplified and without needing double arithmetic:

`{ r / 255.0f, g / 255.0f, b / 255.0f, a / 255.0f }`

loader/include/Geode/utils/ColorCode.hpp
```diff
@@ -0,0 +1,92 @@
...
+    };
+
+    struct GEODE_DLL Color final {
+        /// Parses the given hex string using one of the following formats:
+        /// #R/G/B, #RR/GG/BB, #RGB, #RGBA, #RRGGBB or #RRGGBBAA
+        /// Where the # prefix is optional and / means it's 1 byte/nibble representing all channels
+        /// @returns The parsed result
+        static Result<Color> parse(std::string_view hex);
+    private:
+        inline static GLubyte getByte(uint32_t colorNum, size_t index, bool isShort);
```
dankmeme01: this was probably meant to be just `static` because the method is  not inline

dankmeme01 (changes requested):
Aside from pending comments, there are some other things overall:

* While we aren't very consistent with it, most of the codebase doesn't add `const` to all local variables, so it's best to remove that for clarity
* Simpler functions should honestly just be in the header for consistency. Those with a larger body can stay in the .cpp file. I probs would've just made the entire class header-only so it doesn't add extra compile time to geode and allows for greater inlining

loader/include/Geode/utils/ColorCode.hpp
SMJSProductions: Transparent is see through but not invisible. This was mostly based of some other color classes I've seen before and originated from some niche use I had in GDI

loader/src/utils/ColorCode.cpp
SMJSProductions: It's moreso not continuing operations if you send it "#" or "" because otherwise I do all kinds of checks on something which will fail regardless.

loader/src/utils/ColorCode.cpp
SMJSProductions: Fair enough, this kinda stemmed from its original state

loader/src/utils/ColorCode.cpp
SMJSProductions: It does. GDI uses numFromString and that uses capital colors

PR #2123, opened 2026-07-16
Added a data method to WebRequest to get the body ref rather than a copy

SMJSProductions:


Fleeym:
Not sure if `data()` is the right name here, maybe `getBodyRef()` would be a bit better?

dankmeme01:
Also upon another thought, we should not return a reference to an option because it's generally a bad idea. instead turn it into `std::optional<std::span<const uint8_t>>`

Fleeym:
I bless with LGTM!

PR #2139, opened 2026-07-29
Add public sha256 utils

dankmeme01:
This refactors and removes most of the internal `loader/hash/hash.hpp`, keeping only the very Geode-specific `sha256Text` method that ignores \n and \r characters.

Instead, there are now public SHA256 utils in the `Geode/utils/hash.hpp` header: both a stateful `Sha256Hasher` for incremental hashing and one-shot `sha256` functions. As well as a shared `Hash` class, allowing users to choose whether they want a binary digest or a hex form.


Cvolton (approved):
lgtm

PR #2142, opened 2026-07-30
add CCConfiguration fix

prevter:
cocos2d forgets to reload the extensions string on GL context recreation, which results in a dangling pointer.
the fix is to copy the extensions string to an owned buffer.


Cvolton (approved):
lgtm

PR #2152, opened 2026-08-06
Save optimization

Cvolton:
Draft for now, since I see some space for improvements, but already creating the PR to get a second pair of eyes on this so I can see if this approach is worth pursuing.

This PR supersedes #1952. It adds "dirty" and "taken" flag tracking to saved and settings jsons and avoids saving them. As mentioned before, the motivation is primarily Android saving performance, though this leads to measurable improvements on Windows as well.

The idea is simple - do not resave settings and saved jsons if they haven't changed. To achieve this it essentially tracks 3 states - clean, dirty, taken. Clean is simple - do not save. Dirty means that it should be saved the next time the game saves but it doesn't need to be saved again afterwards. Taken means that a ref to the container may be stored somewhere, so it should be resaved every time until game exit. This logic is applied to both settings and saved values. Dirty flag is also set if the mod contains settings the user does not have or if the settings file is corrupted (didn't successfully load) for some reason.

In addition to this major change, the PR contains the following minor changes:
- Mod settings are now saved when the user clicks "Apply" in the mod settings popup (to save time on game exit)
- Saved values are now loaded even if settings fail to load

Stuff I'd like to figure out before marking this prod ready
- Using RAII to track container borrow lifespans to not require a permanent taken flag (might not be doable without V6 though?)
- probs just that, i also need to write doc comments and maybe reorder the members i added

thanks for reading my big yap

### DIFF
--- a/loader/include/Geode/loader/Mod.hpp
+++ b/loader/include/Geode/loader/Mod.hpp
@@ -307,7 +307,7 @@ namespace geode {
                 return res.unwrap();
             }
 
-            auto savedMutable = this->getSaveContainerTemp();
+            auto& savedMutable = this->getSaveContainerTemp();
             savedMutable[key] = matjson::Value(defaultValue);
             return defaultValue;
         }
--- a/loader/include/Geode/loader/Mod.hpp
+++ b/loader/include/Geode/loader/Mod.hpp
@@ -82,6 +82,8 @@ namespace geode {
         friend void GEODE_CALL ::geode_implicit_load(Mod*);
 
         void settingReact(geode::Function<void()> fn);
+        
+        matjson::Value& getSaveContainerTemp();
     public:
         // no copying
         Mod(Mod const&) = delete;
@@ -242,7 +244,6 @@ namespace geode {
         }
 
         matjson::Value& getSaveContainer();
-        matjson::Value& getSaveContainerTemp();
         matjson::Value const& getSaveContainerConst() const;
         matjson::Value& getSavedSettingsData();
 
--- a/loader/include/Geode/loader/Mod.hpp
+++ b/loader/include/Geode/loader/Mod.hpp
@@ -242,6 +242,7 @@ namespace geode {
         }
 
         matjson::Value& getSaveContainer();
+        matjson::Value const& getSaveContainerConst() const;
         matjson::Value& getSavedSettingsData();
 
         /**
@@ -286,7 +287,7 @@ namespace geode {
 
         template <class T>
         T getSavedValue(std::string_view key) {
-            auto& saved = this->getSaveContainer();
+            auto& saved = this->getSaveContainerConst();
             if (auto res = saved.get(key).andThen([](auto&& v) {
                 return v.template as<T>();
             }); res.isOk()) {
@@ -297,13 +298,15 @@ namespace geode {
 
         template <class T>
         T getSavedValue(std::string_view key, T const& defaultValue) {
-            auto& saved = this->getSaveContainer();
+            auto& saved = this->getSaveContainerConst();
             if (auto res = saved.get(key).andThen([](auto&& v) {
                 return v.template as<T>();
             }); res.isOk()) {
                 return res.unwrap();
             }
-            saved[key] = matjson::Value(defaultValue);
+
+            auto savedMutable = this->getSaveContainer();
+            savedMutable[key] = matjson::Value(defaultValue);
             return defaultValue;
         }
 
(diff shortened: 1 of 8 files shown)


Cvolton:
upon further thought, I decided that using RAII to manage lifetimes here would likely not work as intended, even assuming a breaking change. A wrapper struct could be made, however for proper json interfacing we'd still need to keep the json struct extractable, which fundamentally breaks the contract given by the potential wrapper struct.

For the same reason I've decided to make getSaveContainerTemp private, since there is also no way to guarantee that the caller won't just keep holding onto the container forever. The Const version is fine to stay public from this perspective IMO, since the const return type alone shows that it should be read only and if the caller const_casts it away, the worst thing that can happen is that their changes won't be saved.

2 extremely minor semantical questions remain
- is it better to express the dirty/taken difference using two bools (current version) or an enum?
- the markDirty/unmarkDirty func pair might be better as markDirty(bool)?

With that said I think I'm happy enough with this for now that I can mark this as ready for review.

Fleeym:
I'd say an enum works a bit better for dirty / taken, since the 2 bools exclude each other (it doesn't make sense for both to be true at the same time). Since it's not exposed in the API it should be fine the way it is right now too.

loader/include/Geode/loader/Mod.hpp
```diff
@@ -297,13 +300,15 @@ namespace geode {
...
+            auto& saved = this->getSaveContainerConst();
             if (auto res = saved.get(key).andThen([](auto&& v) {
                 return v.template as<T>();
             }); res.isOk()) {
                 return res.unwrap();
             }
-            saved[key] = matjson::Value(defaultValue);
+
+            auto savedMutable = this->getSaveContainerTemp();
+            savedMutable[key] = matjson::Value(defaultValue);
```
dankmeme01: missing `auto&`, this copies the json into a local var so the change is meaningless

loader/include/Geode/loader/Mod.hpp
```diff
@@ -316,8 +321,22 @@ namespace geode {
...
             auto old = this->getSavedValue<T>(key);
+
+            // optimization: if the value is the same, don't write to the save container
+            // constexpr checks needed to avoid compile errors for types that don't support operator==
+            if constexpr (std::ranges::range<T>) {
+                using Elem = std::ranges::range_value_t<T>;
+                if constexpr (requires(Elem const& a, Elem const& b) { { a == b } -> std::convertible_to<bool>; }) {
+                    if (old == value) return old;
+                }
+            }
```
dankmeme01, line 333: is this entire block necessary? both blocks do the exact same thing - `if (old == value)`, and the block below should work alone since it already checks whether the equation will compile

loader/src/loader/ModImpl.cpp
```diff
@@ -209,11 +227,18 @@ Result<> Mod::Impl::loadData() {
...
     auto settingPath = m_saveDirPath / "settings.json";
     if (std::filesystem::exists(settingPath)) {
-        GEODE_UNWRAP_INTO(auto json, utils::file::readJson(settingPath));
-        auto load = m_settings->load(json);
-        if (!load) {
-            log::warn("Unable to load settings: {}", load.unwrapErr());
+        if (auto json = utils::file::readJson(settingPath)) {
+            auto load = m_settings->load(json.unwrap());
+            if (!load) {
+                log::warn("Unable to load settings: {}", load.unwrapErr());
```
dankmeme01, line 234: should we also `queueSave()` here like in other failure paths?

loader/src/loader/ModImpl.cpp
```diff
@@ -243,18 +268,36 @@ Result<> Mod::Impl::saveData() {
...
+        ModStateEvent(ModEventType::DataSaved, std::move(m_self)).send();
     }
-    auto res2 = utils::file::writeStringSafe(m_saveDirPath / "saved.json", m_saved.dump());
-    if (!res2) {
-        log::error("Unable to save values: {}", res2.unwrapErr());
+
+    if (m_saveRequestState != SaveRequestState::Clean) {
+        log::debug("Saving values for mod {}", m_metadata.getID());
+
+        auto res2 = utils::file::writeStringSafe(m_saveDirPath / "saved.json", m_saved.dump());
```
dankmeme01: could be just `res` since we are in a scope now

loader/include/Geode/loader/Mod.hpp
Cvolton: the one with ranges is necessary because otherwise it fails for containers

dankmeme01: what's the error? the second block checks if `T == T` is valid, I don't see how that's not enough to conclude `old == value` will compile, considering they *are* `T`

Cvolton: it fails with vectors that have children that are not directly comparable - it incorrectly passes the check and then fails to compile when vector itself has operator== but the contained elements don't

loader/include/Geode/loader/Mod.hpp
Cvolton: fixed

PR #2155, opened 2026-08-12
Download loader updates from links given by the server

dankmeme01:
No longer hardcode `github.com/geode-sdk/geode`, instead the server now returns all needed URLs.

Also comes with a hash verification which previously was never a thing, and a small refactor of update installation because of the crazy nesting.


Cvolton:
maybe should handle blank hash? otherwise lgtm

dankmeme01:
Real I forgot about that

PR #2160, opened 2026-08-15
fix a few bugs in numToAbbreviatedString

slideglide:
the function casted to float, which lost precision on large 64 bit values, a few other issue too:

- negative numbers were never abbreviated
- the bucket is picked from the raw int, but the displayed digits come from a separately truncated float; these can disagree right at a threshold boundary

also added Q and T tiers since they were missing for 64 bit values

### DIFF
--- a/loader/include/Geode/utils/general.hpp
+++ b/loader/include/Geode/utils/general.hpp
@@ -152,24 +152,23 @@ namespace geode {
         template <std::integral Num> requires (!std::same_as<std::remove_cv_t<Num>, bool>)
         std::string numToAbbreviatedString(Num num) {
             bool negative = false;
-            using Unsigned = std::make_unsigned_t<std::common_type_t<Num, int64_t>>;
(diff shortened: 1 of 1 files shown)


slideglide:
changed the function to not compile if it gets passed a bool

Cvolton:
> changed the function to not compile if it gets passed a bool

technically an api break if the old impl accepted it? though also i cant imagine anyone passing a bool into this except for maybe accidentally getting `operator bool()`'d so probs acceptable even pre-v6

slideglide:
> > changed the function to not compile if it gets passed a bool
> 
> technically an api break if the old impl accepted it? though also i cant imagine anyone passing a bool into this except for maybe accidentally getting `operator bool()`'d so probs acceptable even pre-v6

i can remove the requirement then, i checked whether any mod did this using github search but i can't seem to find any

Cvolton:
im vibemerging this, the chakras in this pr feel good

PR #2177, opened 2026-09-04
geode::Toggler

slideglide:
A nice toggler button that wraps around geode::Button

Benefits over CCMenuItemToggler:
- Same benefits as geode::Button
- Does not have the inverted logic bug like CCMenuItemToggler

### DIFF
--- a/loader/include/Geode/ui/Button.hpp
+++ b/loader/include/Geode/ui/Button.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <Geode/utils/ZStringView.hpp>
+#include <Geode/utils/function.hpp>
 #include <cocos2d.h>
 
 namespace geode {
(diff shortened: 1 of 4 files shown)


loader/include/Geode/ui/Button.hpp
```diff
@@ -145,7 +146,7 @@ namespace geode {
          */
         virtual void setEnabled(bool enabled);
         virtual bool isEnabled();
-
```
qimiko: why is this file being reformatted?

loader/src/ui/nodes/Button.cpp
```diff
@@ -158,7 +158,9 @@ CCNode* Button::getDisplayNode() {
 }
 
 void Button::setDisplayNode(CCNode* node) {
-    m_impl->m_displayNode->removeFromParent();
+    if (m_impl->m_displayNode) {
```
qimiko: this bug fix should be in a separate pull request unless your toggler for some reason causes `m_displayNode` to be nullptr (which it probably shouldn't?)

loader/src/ui/nodes/Button.cpp
```diff
@@ -297,7 +299,7 @@ void Button::selected() {
 
 void Button::unselected() {
     if (!m_impl->m_selected) return;
-    
+
```
qimiko: like `Button.hpp`, somewhat unnecessary formatting changes

PR #2185, opened 2026-09-10
[v6] migrate most CCLabelBMFont usage to geode::Label

Fleeym:
Tried to not just find & replace, I attempted to optimize create() and setString() usage wherever I could.

This obviously will lead to some API breaks, so I'm basing it on v6.

I think this is the full list:

- `SettingNodeV3::getNameLabel()`: changed return type
- `SettingNodeV3::getStatusLabel()`: changed return type
- `IconButtonSprite::getLabel()`: changed return type
- `Notification::getLabel()`: changed return type
- `Popup::m_title`: is now `geode::Label`
- `ProgressBar::getProgressLabel()`: changed return type
- `SelectList::m_label`: is now `geode::Label`
- `SliderNode::linkLabel()`: requires `geode::Label` for 1st param
- `SliderNode::getLinkedLabel()`: changed return type
- `SimpleTextArea::getLines()`: returns `vector<Label*>`

The majority of those should not be issues for most devs, besides Popup, SliderNode and SelectList.

### DIFF
--- a/loader/src/ui/mods/settings/SettingNodeV3.cpp
+++ b/loader/src/ui/mods/settings/SettingNodeV3.cpp
@@ -5,6 +5,7 @@
 #include <Geode/ui/MDPopup.hpp>
 #include <Geode/ui/Scrollbar.hpp>
 #include <Geode/ui/Button.hpp>
+#include <Geode/ui/Label.hpp>
 #include <Geode/binding/TextArea.hpp>
 #include "KeybindEditPopup.hpp"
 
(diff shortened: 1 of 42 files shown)


loader/src/ui/nodes/IconButtonSprite.cpp
```diff
@@ -50,7 +51,7 @@ void IconButtonSprite::updateLayout() {
 
     CCSize size = ccp(PADDING * 2, 35);
     if (hasText) {
-        m_impl->label->limitLabelWidth(100, .6f, .1f);
+        m_impl->label->setLimitLabelWidth(100, .6f, .1f);
```
dankmeme01: can be moved into `init`, this setting is remembered forever after being set unlike `CCLabelBMFont`

loader/src/ui/mods/settings/SettingNodeV3.cpp
```diff
@@ -636,20 +637,20 @@ void FileSettingNodeV3::updateState(CCNode* invoker) {
...
         m_nameLabel->setOpacity(155);
     }
     else {
-        m_nameLabel->setString(utils::string::pathToString(this->getValue().filename()).c_str());
+        m_nameLabel->setText(utils::string::pathToString(this->getValue().filename()));
         m_nameLabel->setColor(ccWHITE);
         m_nameLabel->setOpacity(255);
     }
-    m_nameLabel->limitLabelWidth(75, .35f, .1f);
+    m_nameLabel->setLimitLabelWidth(75, .35f, .1f);
```
dankmeme01: can also be moved into init

dankmeme01 (approved):

